PR #33, opened 2013-11-21
Default to no compiler optimization for extensions

kcarnold:
clang++ miscompiles the stanc parser at high optimization levels (see https://groups.google.com/d/msg/stan-users/HP3YjPrBKO0/VVuJyE3Gs-EJ).

Fortunately, stanc isn't the thing that needs to be fast.

Since libstan is not affected by this issue, this commit splits out
the compiler flags for libstan from those for the extensions (mostly
stanc), and backs the latter down to -O0.

(Note that I haven't actually tested this exact code, since compiling still takes a while...)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,14 +99,16 @@ libstan_sources = [
     "pystan/stan/src/stan/agrad/matrix.cpp"
 ]
 
-extra_compile_args = ['-O3', '-ftemplate-depth-256']
+libstan_extra_compile_args = ['-O3', '-ftemplate-depth-256']
 
 libstan = ('stan', {'sources': libstan_sources,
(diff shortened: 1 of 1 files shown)


ariddell:
@kcarnold Did you ever try -O2 instead of -O3? The binary size of PyStan (now that we're shipping wheels) is becoming something of an issue.

kcarnold:
I don't remember. Probably some clang bugs have been fixed in the last 3 years :) But if binary size is an issue, I'd suggest trying `-Os` too (which is [apparently](http://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-O0) `-O2` plus additional size optimizations).

PR #141, opened 2015-01-28
added kwargs to traceplot

nesanders:
Meant to enable users to pass additional arguments, such as figsize, to the pymc traceplot function.


ariddell:
LGTM. Thanks Nathan.

Issue #3, opened 2013-08-16
PyStan broken on IPython Notebook + qtconsole ('OutStream' error)


sjobeek:
Running into the following error on Windows (32 bit Anaconda installation) when I attempt to run the example:

```
AttributeError: 'OutStream' object has no attribute 'fileno'
```

I read that this issue can be related to the IPython Notebook and how it uses /over-writes the stderr

See what looked to be a similar issue several months ago in SciPy: 
https://github.com/ipython/ipython/issues/3017

I can confirm that this issue happens in the IPython Notebook and the qtconsole. The same code is tested and working fine from the terminal. 

Since the IPython notebook is so popular, this is definitely one to fix!

Thanks for implementing PyStan, though! This is going to be quite useful =)

```
WARNING:root:COMPILING THE C++ CODE FOR MODEL anon_model NOW.
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-5-f3d4d4964bda> in <module>()
      1 fit = pystan.stan(model_code=schools_code, data=schools_dat,
----> 2                   iter=1000, chains=4)

C:\Anaconda\lib\site-packages\pystan\api.pyc in stan(file, model_name, model_code, fit, data, pars, chains, iter, warmup, thin, init, seed, sample_file, diagnostic_file, save_dso, verbose, boost_lib, eigen_lib, **kwargs)
    246         m = StanModel(file=file, model_name=model_name, model_code=model_code,
    247                       boost_lib=boost_lib, eigen_lib=eigen_lib,
--> 248                       save_dso=save_dso, verbose=verbose, **kwargs)
    249     if sample_file is not None:
    250         raise NotImplementedError

C:\Anaconda\lib\site-packages\pystan\model.pyc in __init__(self, file, charset, model_name, model_code, stanc_ret, boost_lib, eigen_lib, save_dso, verbose, **kwargs)
    258         if not verbose:
    259             # silence stderr for compilation
--> 260             orig_stderr = pystan.misc.redirect_stderr()
    261 
    262         build_extension.run()

C:\Anaconda\lib\site-packages\pystan\misc.py in redirect_stderr()
    438     """
    439     sys.stderr.flush()
--> 440     stderr_fileno = sys.stderr.fileno()
    441     orig_stderr = os.dup(stderr_fileno)
    442     devnull = os.open(os.devnull, os.O_WRONLY)

AttributeError: 'OutStream' object has no attribute 'fileno'

```

[...]

PR #144, opened 2015-02-12
misc._has_fileno should check for IOError also

dougalsutherland:
Apparently that's what Python 2 raises: https://docs.python.org/2/library/io.html#io.IOBase.fileno (whereas 3 does OSError).

Per @JanSchulz's comment on https://github.com/stan-dev/pystan/issues/3.


ariddell:
Thanks!

PR #166, opened 2015-08-03
Fix doc about windows install

icoxfog417:
Add more specific description about windows install.
Especially, fix wrong option (not `Ehsc`, but `EHsc`).


ariddell:
This is wonderful. Thanks!

PR #167, opened 2015-08-03
first draft of installation instructions

chendaniely:


chendaniely:
Before this get's merged.  It seems I should move the installation_beginner.md file into docs?
I don't have a windows machine at the moment to fully test the windows installation, I originally thought it would work out of the box, but from #166 the windows installation is a bit more complicated?

Should we combine it all into 1 document? or leave the windows part separate?

ariddell:
This is great. Thanks!

Now that I look at the existing docs, perhaps it would be easiest to create a new rst page `installation.rst` and put the details there. In the README we could link to that page. The page that needs to be split up is this one: https://pystan.readthedocs.org/en/latest/getting_started.html#windows which is found in doc/getting_started.rst.

I do like putting the reference to the detailed instructions before the quickstart.

ariddell:
I'd like to merge this. I think it would be good to add some indication that it's possible to install things without Anaconda on OS X (since numpy has binary wheels available) and Linux without too much fuss. Clearly Anaconda is needed (and Python 3.5) on Windows until NumPy has binary wheels available (although this looks like it could happen within weeks).

Any thoughts @chendaniely ?

chendaniely:
@ariddell added comment/section about anaconda not needed to be an absolute requirement.

Otherwise feel free to merge.

ariddell:
Looks good. Thanks!

PR #172, opened 2015-08-21
Make testing more forgiving

tobeplugged:
This uses less strict comparison for numerical results to account for rounding errors.
I've also removed two unneeded imports from the file.

If you want, I can squash those commits.

### DIFF
--- a/pystan/tests/test_rstan_stanfit.py
+++ b/pystan/tests/test_rstan_stanfit.py
@@ -1,5 +1,3 @@
-import os
-import tempfile
 import unittest
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


bob-carpenter:
Thanks --- that looks good, though I don't know the Python constructs.  You almost always want to test floating point to within a tolerance rather than exactly because different compilers implement arithmetic differently --- the specs only require a minimum accuracy and then compilers like the Intel ones don't even follow the spec's minimum accuracy requirements (though they do offer a slower, conforming mode).

tobeplugged:
Yeah as soon as I started to use testing I got bitten by that.

ariddell:
Thanks!

PR #187, opened 2015-11-04
Feature/update to stan develop

syclik:
This brings the Stan branch up to the current develop branch which includes updates to the writer callbacks. 

@ariddell, let me know if you want more doc on the C++ side. Also, if you want to wait to merge this until later, that's fine and I'll keep updating the branch with additional changes to the Stan library.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ __pycache__
 *.egg-info
 
 .tox/
+
+*~
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -16,8 +16,8 @@ prune pystan/stan/src/python
 prune pystan/stan/src/test
 
 # unused parts of the stan math source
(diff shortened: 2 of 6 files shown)


ariddell:
I'm going to try to do a point release and then merge this.

Is there a reason why the filtered values / sum values code is moving out of Stan? Is it going to be identical/duplicated in RStan?

syclik:
Sounds good. There's really no rush and if you want me to close this and just leave it as a running branch until you want it, let me know. I'm happy doing that.

Re: values... good question. It might get put back in after the dust settles. It will be duplicated for RStan and for the time being, we can duplicate. We should be able to move an implementation back in once we've got final places for everything.

ariddell:
LGTM! Thanks for all this work on the refactor (to @betanalpha as well).

syclik:
Thanks. I'll keep the branch pretty up-to-date with Stan so there's less
work to do later.

PR #192, opened 2015-11-10
Remove over-wide assert from _chains

### LINKED ISSUE #191: `assert` in `_chains.effective_sample_size` triggers on constant parameters
I'm coming across instances where the `assert chain_var[chain] < float('inf'), chain_var[chain]` line in _chains.pyx triggers when I'd rather it didn't. I have a model with a transformed parameter which is the same on every step of the chain. It has standard deviation zero, and so the autocovariance is nan. This causes an `AssertionError` in `_chains.effective_sample_size`.

@ariddell , I know I suggested in #186 checking that `chain_var[chain]` isn't infinity, but it seems there's cases where that should pass. For this particular case, the `Rhat` in the summary output shows up as `NaN` (if I remove the `assert`), so that makes it obvious that the effective sample size output is suspect.

stephen-hoover:
In `_chains.effective_sample_size`, remove the assertion that the variance is finite. For some infinities, it's better to let the code continue and have the user decide what to do with the output.

Closes #191 .

### DIFF
--- a/pystan/_chains.pyx
+++ b/pystan/_chains.pyx
@@ -125,7 +125,6 @@ def effective_sample_size(dict sim, int n):
         n_kept_samples = ns_kept[chain]
         chain_mean.push_back(get_chain_mean(sim, chain, n))
         chain_var.push_back(acov[chain][0] * n_kept_samples / (n_kept_samples-1))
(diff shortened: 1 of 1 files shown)


ariddell:
Thanks for catching this.

Issue #204, opened 2015-12-27
Cache StanProgram/StanModel compilation


ariddell:
PyStan should cache [StanProgram](https://github.com/stan-dev/stan/wiki/User-Interface-Guidelines-for-Developers) (née StanModel) compilation by default. Compiled extension
objects are small (ca. 5M) and model compilation takes a great deal of time. Excellent work on this has been done by Paul Kernfeld (@paulkernfeld) (https://github.com/paulkernfeld/pystan-cache).

Desiderata:
- [ ] Use standard cache locations for OS X, Linux, and Windows; this is default for `cache_dir`
- [ ] Cache by default, `cache=False` will disable
- [ ] Cache 350M worth of models by default using a LRU cache

(I picked 350M because that's the default amount of disk space Firefox
will use for caching images if you disable automatic cache management.)

Details:
- [ ] Hash of Stan model code + Stan version + platform works as a unique identifier

ariddell:
I'd love to have this feature. What about just starting with an experimental version which doesn't have the LRU cache and doesn't pay attention to disk space constraints?

ariddell:
For what it's worth, I've been informally caching my Stan models and have accumulated 4G in 4 months. I think a default cache of 500M is reasonable.

bob-carpenter:
Seems reasonable to me (starting simple and 500MB cache size).

Is there a way to use the temp directory so we don't drop 500MB
of data onto someone's drive without them realizing it and without
it getting autocleaned?  Otherwise, you may want to have them
explicitly turn it on and size it so they know what they're getting
into.

I think many people are going to be working through lots of
model variants.

- Bob

ariddell:
We're using the standard cache directories (OS X, Windows, Linux). They're known to the operating system as temporary.

ariddell:
Note that if you compress a model using LZMA (part of Python standard library), it only takes ~3M rather than ~25M.

ariddell:
PyStan 3 has this.

sursu:
@ariddell in case you know: what is the expected release date of PyStan 3?

riddell-stan:
There's a working alpha version which you can install with `python3 -m pip install --pre pystan`. See https://github.com/stan-dev/pystan-next

sursu:
True, but I'm reluctant to use it when it says "Do not use!".
An estimated release date would be nice.

riddell-stan:
[...]

PR #237, opened 2016-07-07
add a temp_dir variable to allow storing compiled models in other places

segasai:
#### Summary:

Currently the compiled models (.so files) are stored in /tmp/ and deleted immediately after use. However the files are kept open by a process and occupy disk space. Therefore if in one python session many models are compiled, the amount of disk space consumed could be many gigabytes. That could lead to exhaustion of space in /tmp/. Therefore it makes sense to allow the compiled models to go in places other than /tmp.
#### Intended Effect:

Add temp_dir option to StanModel
#### How to Verify:

code = 'data{real x;}parameters{real y;}model{ x~normal(y,1);}'
model = pystan.StanModel(StringIO.StringIO(code), temp_dir=__PATH_TO_OTHER_DIR)
#### Side Effects:

None 
#### Documentation:

Added
#### Copyright and Licensing

Myself

### DIFF
--- a/pystan/model.py
+++ b/pystan/model.py
@@ -248,6 +248,7 @@ class StanModel:
         nonce = abs(hash((self.model_name, time.time())))
         self.module_name = 'stanfit4{}_{}'.format(self.model_name, nonce)
         lib_dir = tempfile.mkdtemp(dir=temp_dir)
+        self.temp_dir = temp_dir
         pystan_dir = os.path.dirname(__file__)
         include_dirs = [
             lib_dir,
@@ -364,7 +365,7 @@ class StanModel:
 
     def __setstate__(self, state):
         self.__dict__.update(state)
-        lib_dir = tempfile.mkdtemp(dir=temp_dir)
+        lib_dir = tempfile.mkdtemp(dir=self.temp_dir)
         with io.open(os.path.join(lib_dir, self.module_filename), 'wb') as f:
             f.write(self.module_bytes)
         try:
--- a/pystan/model.py
+++ b/pystan/model.py
@@ -129,6 +129,11 @@ class StanModel:
     eigen_lib : string
         The path to a version of the Eigen C++ library to use instead of
         the one in the supplied with PyStan.
+    
+    temp_dir : string
+        The path to the temporary directory where compiled models are
+        stored during the execution. By default it is the directory
+        specified by tempfile.tempdir variable.
 
     verbose : boolean, False by default
         Indicates whether intermediate output should be piped to the console.
@@ -200,7 +205,8 @@ class StanModel:
     """
     def __init__(self, file=None, charset='utf-8', model_name="anon_model",
                  model_code=None, stanc_ret=None, boost_lib=None,
-                 eigen_lib=None, verbose=False, obfuscate_model_name=True):
+                 eigen_lib=None, verbose=False, obfuscate_model_name=True,
+                 temp_dir=None):
 
         if stanc_ret is None:
             stanc_ret = pystan.api.stanc(file=file,
@@ -241,7 +247,7 @@ class StanModel:
         # module_name needs to be unique so that each model instance has its own module
         nonce = abs(hash((self.model_name, time.time())))
         self.module_name = 'stanfit4{}_{}'.format(self.model_name, nonce)
-        lib_dir = tempfile.mkdtemp()
+        lib_dir = tempfile.mkdtemp(dir=temp_dir)
         pystan_dir = os.path.dirname(__file__)
         include_dirs = [
             lib_dir,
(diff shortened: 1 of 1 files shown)


ariddell:
Thanks for this @segasai. The way things are currently setup the situation is somewhat different than you describe, I think. To allow for consistent pickling of models the compiled .so file is read into *memory* and deleted from disk, see https://github.com/stan-dev/pystan/blob/develop/pystan/model.py#L323 .

If your model is ``m`` you should be able to ``del m`` and get rid of all traces of it. Does that solve your problem?

segasai:
I have seen the code  https://github.com/stan-dev/pystan/blob/develop/pystan/model.py#L323 however I think the problem is that the file is still loaded first here as a module https://github.com/stan-dev/pystan/blob/develop/pystan/model.py#L318 and that is not being cleaned up. For example with this test program and simplest model
`data  {  real x; } parameters   {  real y; } model {  x~normal(y,1); }`
 I managed to fill ~ 1GB of disk space:

```
import pystan

nit =100
code = open('start.stan').read()
for i in range(nit):
	M = pystan.StanModel(model_code=code)
	del M	
```
And I can see how the files are being kept open:
```
lsof -p 29106|grep tmp
python  29106 koposov  DEL    REG    8,2              164877 /tmp/tmpwigxOU/stanfit4anon_model_5b549a88ee01aa6d153f2a577913f192_8380356191924424984.so
python  29106 koposov  DEL    REG    8,2              164874 /tmp/tmpMP7jC6/stanfit4anon_model_5b549a88ee01aa6d153f2a577913f192_8383078519410775059.so
python  29106 koposov  DEL    REG    8,2              164873 /tmp/tmpr0nUvK/stanfit4anon_model_5b549a88ee01aa6d153f2a577913f192_8382959730803202559.so
python  29106 koposov  DEL    REG    8,2              164872 /tmp/tmpU2aVIy/stanfit4anon_model_5b549a88ee01aa6d153f2a577913f192_8380371072972332709.so
python  29106 koposov  DEL    REG    8,2              164867 /tmp/tmpBaOcM9/stanfit4anon_model_5b549a88ee01aa6d153f2a577913f192_8380352492719994984.so
python  29106 koposov  DEL    REG    8,2              164864 /tmp/tmpfwLUzu/stanfit4anon_model_5b549a88ee01aa6d153f2a577913f192_8380443573700740759.so
python  29106 koposov  DEL    REG    8,2              164565 /tmp/tmpmGswpG/stanfit4anon_model_5b549a88ee01aa6d153f2a577913f192_8381461657770409434.so
```

I also tried to force garbage collection by doing gc.collect() -- that doesn't help. 
A note: I usually don't try to compile the same model again and again (I use caching based on md5 of the code), but when you change the model, the same issue occurs.

segasai:
Not quite, the files are gone from the filesystem (i.e. they are invisible), but because the filehandles on them are kept open, they still occupy the space. 
Actually I tried to hack model.py by putting 
`open(os.path.join(lib_dir, self.module_filename), 'w').close()`
before rmtree (this will essentially zero out the file) and got segfault, so you can't do that -- i.e. you *have* to keep the contents of the file intact.
I also tried in load_module() function to explicitely close the file handle returned by find_module but it doesn't seem to help.

I even tried to put this destructor and empty the file during the destruction of StanModel:
```
def __del__(self):
        print( 'destructed')
        open(os.path.join(self.lib_dir, self.module_filename), 'w').close()
        shutil.rmtree(self.lib_dir, ignore_errors=True)
```
However it also crashes. I think the underlying reason is here http://stackoverflow.com/questions/8781257/remove-an-imported-python-module -- python cannot unload modules! and while the module is loaded there is no way to free the disk space. Maybe one can reload the module from different file, to close the file...

ariddell:
This is really interesting. Thanks for digging into it. I didn't know
there was no way to unload a module in Python.

Some of the underlying plumbing connecting PyStan with Stan is currently
being redesigned and I think this particular problem will be solved by
the redesign.

ariddell:
@segasai sorry, I lost track of this one. Am I correct in thinking this is mainly a workaround for the unloadable/undeletable module problem?

ariddell:
Related to #204

segasai:
@ariddell Yes, it is correct to say that it is a workaround the inability to properly unload the module and delete the file containing it.

ariddell:
I think I found a better way. You should be able to set the ``TEMPDIR`` environment variable when launching python.  https://docs.python.org/3.5/library/tempfile.html has details.

Let me know if this doesn't work.

jbao:
@ariddell True, I used the same approach to fix exactly the same issue, it took me a while to figure out though, maybe we should add this hint into the documentation?

ariddell:
Good idea, @jbao. Would you like to add a note to the docs via a pull request?

(I'm going to close this particular PR for now since the ``TEMPDIR`` env var does fix the problem.)

jbao:
hmm, I tried to add some docs, but `sphinx` doesn't seem to work. When I did `make html`, I got the following error:

>Exception occurred:
>  File "/home/jbao/miniconda3/envs/py27/lib/python2.7/site-packages/scipy/__init__.py", line 69, in ><module>
>    from numpy import *
>TypeError: 'type' object does not support indexing

Any ideas?

ariddell:
Good question. I tend to do ``sphinx-build doc /tmp/build`` to test things.

PR #255, opened 2016-09-13
TST: Remove test depending on fixed Stan version

ariddell:
Samples drawn using Stan are not guaranteed to be the same across Stan
versions. This test incorrectly make such an assumption.


ariddell:
Incorporated in a different PR

PR #256, opened 2016-09-14
Turn on C++11, compile with -std=c++11

ariddell:
Simple change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 # language is set to c to work around travis ci bug
 # https://github.com/travis-ci/travis-ci/issues/2312
 language: c
-sudo: false
+sudo: required
+dist: trusty
 os:
   - linux
 git:
--- a/pystan/model.py
+++ b/pystan/model.py
@@ -278,6 +278,7 @@ class StanModel:
         if extra_compile_args is None:
             extra_compile_args = [
                 '-O2',
+                '-std=c++11',
                 '-ftemplate-depth-256',
                 '-Wno-unused-function',
                 '-Wno-uninitialized',
--- a/setup.py
+++ b/setup.py
@@ -111,6 +111,7 @@ stan_macros = [
 ]
 extra_compile_args = [
     '-Os',
+    '-std=c++11',
     '-ftemplate-depth-256',
     '-Wno-unused-function',
     '-Wno-uninitialized',


ariddell:
gcc 4.6.3 does not understand ``-std=c++11`` but does understand ``-std=c++0x``. List of supported features are here: https://gcc.gnu.org/gcc-4.6/cxx0x_status.html

rtrangucci:
Do we need to worry about gcc 4.6.3? I thought RTools 3.4 dropped support
for that toolchain:

https://cran.r-project.org/bin/windows/Rtools/


Rob

ariddell:
Travis-ci still uses it.

bgoodri:
Can `-std` be set to `c++0x` for g++-4.6 and `c++11` for g++4.9?

ariddell:
Looks like you can get some information about the compiler and version being used. Not sure if it's very robust: http://stackoverflow.com/questions/724664/python-distutils-how-to-get-a-compiler-that-is-going-to-be-used#5192738 and http://stackoverflow.com/questions/5986256/python-distutils-gcc-path?noredirect=1&lq=1 talk about it.

ariddell:
Looks like you can turn on ubuntu trusty (with a more recent gcc) in travis-ci. https://docs.travis-ci.com/user/trusty-ci-environment

But I'm still confused about the benefit here. If it compiles using gcc 4.6 now why break that? What does adding the flag do? Are there compiler optimizations that are made? @bob-carpenter @syclik

bgoodri:
I don't think we want to break under 4.6, so we need to be able to figure
out how to specify `--std` conditionally. This is handled automatically on
CRAN via an environmental variable but I don't know about Python.

ariddell:
Seems like PyStan and RStan should turn this on at the same time to avoid confusing people.

ariddell:
@bob-carpenter says this goes in after 2.16. See #353 .

ahartikainen:
Did CmdStan (/RStan) turn this on or are they putting it on 2.18? Stan / Math support this based on the release notes.

https://github.com/stan-dev/stan/releases/tag/v2.17.0

https://github.com/stan-dev/math/releases/tag/v2.17.0

bob-carpenter:
2.17 was supposed to be the final pre-C++11 release.

bgoodri:
There was some stuff merged into the Makefiles that facilitated C++11/4,
but I think no syntax that is not supported by C++03 is being used yet.

ariddell:
If there's no compelling reason to add this flag for 2.17 then let's not add it.

PR #266, opened 2016-10-22
Use pyximport to remove deprecated importlib function for Py3

### LINKED ISSUE #231: importlib.find_loader is deprecated
Use importlib.util.find_spec instead. Appears on line model.py#50.

jrings:
#### Summary:

Resolves #231 
Uses pyximport to make importable the "normal" way

#### Intended Effect:

Get rid of `importlib`, which is deprecated. Originally the idea was to replace with `find_spec`, but that doesn't work with pyx. But the `pyximport` module that's installed with Cython makes pyx modules importable. I thought it'd work for both Python 2 and 3, but I got compilation errors for Python 2, so I just changed it for Python 3. Also, moved the relevant imports into the `if...else` and replaced the clause with a version check for Python 3.

#### How to Verify:

Running the example in the README worked, did it only on Linux though

#### Side Effects:

None

#### Copyright and Licensing

### DIFF
--- a/pystan/model.py
+++ b/pystan/model.py
@@ -12,8 +12,6 @@ if PY2:
 else:
     from collections.abc import Callable, Iterable
 import datetime
-import importlib
-import imp
 import io
 import itertools
 import logging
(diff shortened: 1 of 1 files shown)


ariddell:
Thanks! This is extremely helpful. I didn't understand exactly what changed about ``importlib``.

Issue #311, opened 2017-03-10
pip install fails with FileNotFoundError: [Errno 2] 


fcitterio:
**Summary**
I got the following error when executing `pip install pystan`:

> FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\citterio\\AppData\\Local\\Programs\\Python\\Python35\\Lib\\site-packages\\pystan\\stan\\lib\\stan_math_2.14.0\\doc\\api\\html\\d0\\d88\\struct_eigen_1_1internal_1_1general__matrix__vector__product_3_01_index_00_01stan_1_1math_1_1var15f7d538e8be7d782eccb613a7b1e9bd.html'

**Description**
The prerequisites of numpy and cython are statisfied and I have installed Visual C++ Build Tools 2015.
The problem is that a file is missing and I can't find it in the right directory:
Actually the list of files in '`C:\\Users\\citterio\\AppData\\Local\\Programs\\Python\\Python35\\Lib\\site-packages\\pystan\\stan\\lib\\stan_math_2.14.0\\doc\\api\\html\\d0\\d88`  is:
![cattura](https://cloud.githubusercontent.com/assets/26324543/23792908/2f37d4ca-0589-11e7-9246-7d7e181e6c37.PNG)

#### Current Output:
The complete command output is 

> Collecting pystan
>   Using cached pystan-2.14.0.0-cp35-cp35m-win_amd64.whl
> Requirement already satisfied: Cython!=0.25.1,>=0.22 in c:\users\citterio\appdata\local\programs\python\python35\lib\site-packages (from pystan)
> Requirement already satisfied: numpy>=1.7 in c:\users\citterio\appdata\local\programs\python\python35\lib\site-packages (from pystan)
> Installing collected packages: pystan
> 
> Exception:
> Traceback (most recent call last):
>   File "C:\Users\citterio\AppData\Local\Programs\Python\Python35\lib\site-packages\pip\basecommand.py", line 215, in main
>     status = self.run(options, args)
>   File "C:\Users\citterio\AppData\Local\Programs\Python\Python35\lib\site-packages\pip\commands\install.py", line 342, in run
>     prefix=options.prefix_path,
>   File "C:\Users\citterio\AppData\Local\Programs\Python\Python35\lib\site-packages\pip\req\req_set.py", line 784, in install
>     **kwargs
[...]

PR #312, opened 2017-03-11
INST: exclude stan_math doc from source/wheel dist

ariddell:
Minor update to MANIFEST.in. stan_math docs were not being excluded from the source distribution and the wheel.

Thanks to @fcitterio for the report in #311.


ariddell:
Single test failed due to travis timeout.

PR #333, opened 2017-05-01
Handle unsigned integers in data

### LINKED ISSUE #303: _split_data doesn't handle unsigned integers
#### Summary:
In [`misc._split_data`](https://github.com/stan-dev/pystan/blob/develop/pystan/misc.py#L361), all input data are checked to make sure they're of `int` or `float` type and encoded appropriately. But in numpy's dtype hierarchy, `uint` and `int` are distinct, so unsigned integers are ignored, and PyStan fails when one or more inputs has a dtype like `uint8`. The issue arises here:

```python
if np.issubdtype(np.asarray(v).dtype, int):
    data_i.update({k.encode('utf-8'): np.asarray(v, dtype=int)})
```

Is there any reason not to check against both `int` and `uint`, since the `dtype` of the input data is immediately set to `int` anyway?

This isn't a big deal in my particular use case, since it's trivial to convert the dtype of the array before passing it to PyStan, but since other users may run into this at some point, it might be worth fixing (assuming there's no potential problem I'm missing). I'm happy to open a PR if the suggested solution makes sense.

ariddell:
_split_data now handles unsigned integers (e.g., numpy.uint8)
by converting them into integers.

Thanks to @tyarkoni for the report and @ahartikainen for the suggested
fix.

Closes #303

### DIFF
--- a/pystan/misc.py
+++ b/pystan/misc.py
@@ -359,7 +359,7 @@ def _split_data(data):
     # map<string, pair<vector<int>, vector<size_t>>> so prepare
     # them accordingly.
     for k, v in data.items():
-        if np.issubdtype(np.asarray(v).dtype, int):
+        if np.issubdtype(np.asarray(v).dtype, np.integer):
             data_i.update({k.encode('utf-8'): np.asarray(v, dtype=int)})
         elif np.issubdtype(np.asarray(v).dtype, float):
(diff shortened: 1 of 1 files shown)


ariddell:
CI tests pass, hit a timeout.

Issue #330, opened 2017-04-24
Discourage use of `pystan.stan` function in docs.


ariddell:
Users should be encouraged to create a StanModel (soon StanProgram) instance before using the ``sampling`` or ``optimzing`` etc methods.

The function ``pystan.stan`` should be replaced wherever it occurs in the docs, esp in the getting started vignette. 

See #327 for additional background.

ariddell:
Fixed in #381

PR #336, opened 2017-05-09
Remove pystan.stan from README.rst and other places in documentation

randommm:
From what is proposed on issue #330, removinh `pystan.stan` from pystan Github repository homepage would be a good start.

#### Copyright and Licensing

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -100,8 +100,8 @@ Example
                    'y': [28,  8, -3,  7, -1,  1, 18, 12],
                    'sigma': [15, 10, 16, 11,  9, 11, 10, 18]}
 
-    fit = pystan.stan(model_code=schools_code, data=schools_dat,
-                      iter=1000, chains=4)
+    sm = pystan.StanModel(model_code=schools_code)
+    fit = sm.sampling(data=schools_dat, iter=1000, chains=4)
 
     print(fit)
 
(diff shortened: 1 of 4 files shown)


ahartikainen:
Do we have any opinions on naming?

    sprogram
    program
    dprogram

randommm:
Good point. I thought of leaving just `sm`, but since he said that will 
be called Stan Program in future so I tryed to predict the naming. Also 
thought of helping with the documentation, but there is a whole section 
about avoiding model recompilation which I don't how should be handled 
since it loses its meaning once `stan.fit` is out of the game on the 
previous sections.

ariddell:
Let's keep it "model" for now, otherwise it's a bit confusing. I think we want to do the 's/model/program/' all at once and across all interfaces. Hopefully it will be done when we harmonize variable names and everything else.

randommm:
OK, I removed all the references to `pystan.stan` I could find, trying to adapt the text as most as possible. Should the API documentation of it also include a deprecation note?

ariddell:
Great. Thanks for this!

ariddell:
I'm not sure about the deprecation schedule. I definitely know that we want to discourage the use since we want to separate models and fits for all sorts of reasons (e.g., they have different methods, they need to be pickled separately, ...)

Issue #321, opened 2017-04-10
implement documentation lookup function


ljwolf:
#### Summary:
When using pystan, it'd be nice to have an equivalent to `rstan`'s [`lookup` ](https://rdrr.io/cran/rstan/man/lookup.html)

#### Description:
It's kind of annoying to have to continuously refer to the pdf reference manual, and would be nice to have a way to access the documentation directly from pystan. This would help improve package usability significantly. 

#### Reproducible Steps:
there is no `pystan.lookup` function. 

#### Current Output:
`''`

#### Expected Output:
feature parity with rstan's lookup function,

#### PyStan Version:
2.14

#### Python Version:
python 3

#### Operating System:
linux

ariddell:
Oh this is rather neat. Seems like this might be equally/more useful as a webpage. Perhaps something like this: http://hyperpolyglot.org/numerical-analysis but with "Stan", "R", "NumPy" columns?

bgoodri:
It is literally a lookup table, so just grab from

https://github.com/stan-dev/rstan/tree/develop/rstan/rstan/tools

ahartikainen:
This is fixed by https://github.com/stan-dev/pystan/pull/337

Closing this issue.

PR #337, opened 2017-05-10
Feature/issue 321 lookuptable

randommm:
#### Summary:
Implement lookup function with automatic importation of R's lookup table and some entries for Python. For now just a few Python entries... later I'll add some more, but I'd rather sent this PR now so I maybe I can get some early feedback.

#### Intended Effect:
Fix issue #321 

#### How to Verify:
Check code, try function and unit tests included.

#### Side Effects:
A little bit of extra memory allocated once the lookup table is built (first usage). This is done to avoid rebuilding it on every usage.

#### Documentation:
Included with the function.

#### Copyright and Licensing


ariddell:
This is really great! Thanks.

Two suggestions to start:
- 4 space indents (PyStan's unofficial style guide is Khan's https://github.com/Khan/style-guides/blob/master/style/python.md)
- When you have data (e.g., the .txt files) shipped with the package one typically needs to use pkg_resources because Python needs to support loading data out of wheels and other compressed binary objects. It's not bad. Here's the API: https://setuptools.readthedocs.io/en/latest/pkg_resources.html?highlight=data#resourcemanager-api I'm fairly sure this situation hasn't changed in the last year or two.

ariddell:
Looks like there might be some python 2 issues (see the CI build log). Also, pandas isn't a requirement for pystan. Perhaps there's a workaround?

ahartikainen:
Pandas could be (unofficial) optional import? Isn't the function returning a dictionary (should it be `OrderedDict`?) if pandas is not found?

ariddell:
Sounds good to me. Something similar is done in the plotting code (when
matplotlib is not found).

If pandas is really important then you might consider just raising a
RuntimeError telling the individual to install pandas (as we do with
matplotlib).

The reason for avoiding pandas as a dependency is that there are people
who want to run pystan on servers and don't need to be bothered with
downloading matplotlib or pandas (or particular versions of them).

randommm:
On 12/05/2017 17:31, Allen Riddell wrote:
>
> Looks like there might be some python 2 issues (see the CI build log).
>
OK, I'll take a look. I completely forgot to test it with python 2.



On 12/05/2017 17:48, Ari Hartikainen wrote:

> Isn't the function returning a dictionary (should it be 
> |OrderedDict|?) if pandas is not found?
>
It actually returns a structured numpy array if pandas is not available. 
If it is, then it returns a pandas dataframe, I did it because pandas 
are way better to read (for example, have title for each column).

I think that maybe the best thing is to have another parameter called 
use_pandas. If use_pandas==True it behaves exactly like right now (use 
pandas if available, and gracefully numpy array if not available), and 
if use_pandas==False, it always return a numpy array (good if someone 
wants to be sure of the result type to use it somewhere). It should 
defaults to True since almost always this function is meant to be used 
iteratively and users want pretty and readable results in that case.

ariddell:
Thanks again for your work on this.

This use_pandas business is a bit confusing, I think. If the person has use_pandas=True but doesn't have pandas installed, it gives a different result than if the person has use_pandas=True but does have pandas installed.

If we're sure that most people are going to have pandas installed let's just return the pandas dataframe or raise an ImportError like we do with plotting (when matplotlib is not available). We can add to the docs that you need to have pandas installed.

Something similar to this:

```
    try:
        from pystan.external.pymc import plots
    except ImportError:
        logger.critical("matplotlib required for plotting.")
        raise
```
except we don't need the logging error.

randommm:
On 15/05/2017 13:14, Allen Riddell wrote:
>
> Thanks again for your work on this.
>
> This use_pandas business is a bit confusing, I think. If the person 
> has use_pandas=True but doesn't have pandas installed, it gives a 
> different result than if the person has use_pandas=True but does have 
> pandas installed.
>
> If we're sure that most people are going to have pandas installed 
> let's just return the pandas dataframe or raise an ImportError like we 
> do with plotting (when matplotlib is not available). We can add to the 
> docs that you need to have pandas installed.
>
OK, so is it better to get rid of use_pandas and always raise 
ImportError if it pandas is not available; or should it be kept and 
change the behavior to something more concise? That is:

     use_pandas : bool
         If True and `pandas` is available, will return a `pandas`
         object.
         If True and `pandas` is not available, will raise ImportError.
         If False, will return a structured `numpy` array.
         Defaults to True.

I'm happy with either solution, and anyway the first is forward 
compatible with the second (that is, we can add parameter use_pandas in 
the future if needed).


> Something similar to this:
>
> |try: from pystan.external.pymc import plots except ImportError: 
> logger.critical("matplotlib required for plotting.") raise |
>
> except we don't need the logging error.
>
Why it's not necessary?

ariddell:
Background: I'm just following the principle of least surprise here. Pandas, for example, raises an import error (or something like it) when you use pandas.read_excel but don't have whatever extra package is needed (xlrd, I think). So I think the pattern of raising an import error when a required package is missing is an established one.

About logging, I said don't worry about it because it's of limited value in the interactive setting. The jupyter notebook might swallow it completely, so the user never sees it. In fact, I think the way I wrote the error above is wrong for this reason. The error message should be in the ImportError's message.

This is a better model which is closer to how pandas does things:
```
    try:
        from pystan.external.pymc import plots
    except ImportError:
        raise ImportError('pystan plotting functions require matplotlib.')
```

randommm:
OK, got rid of `use_pandas`.

ariddell:
Looks good to me. I think you just need to update the tests. (Maybe add pandas to the test requirements if it isn't already there?)

PR #359, opened 2017-06-25
Enable vars argument in misc._print_stanfit

ahartikainen:
Removed `if-else` block and moved the `pars` handling to `_summary` function.
Now one can print summary for chosen pars. 

`print(pystan.misc._print_stanfit(fit, pars=['alpha', 'beta[0]', 'sigma'], probs=(0.025, 0.50, 0.975), digits_summary=3))`

#### Summary:

Removed if-else block in _print_stanfit function. It did nothing but disabled the pars argument.

#### Intended Effect:

One can call chosen pars.

#### How to Verify:

`print(pystan.misc._print_stanfit(fit, pars=['alpha', 'beta[0]', 'sigma'], probs=(0.025, 0.50, 0.975), digits_summary=3))`

#### Side Effects:

None

#### Copyright and Licensing

### DIFF
--- a/pystan/__init__.py
+++ b/pystan/__init__.py
@@ -7,7 +7,7 @@
 import logging
 
 from pystan.api import stanc, stan
-from pystan.misc import read_rdump, stan_rdump
+from pystan.misc import read_rdump, stan_rdump, stansummary
 from pystan.model import StanModel
 from pystan.lookup import lookup
 
--- a/pystan/misc.py
+++ b/pystan/misc.py
@@ -50,7 +50,7 @@ from pystan.constants import (MAX_UINT, sampling_algo_t, optim_algo_t,
 logger = logging.getLogger('pystan')
 
 
-def _print_stanfit(fit, pars=None, probs=(0.025, 0.25, 0.5, 0.75, 0.975), digits_summary=2):
+def stansummary(fit, pars=None, probs=(0.025, 0.25, 0.5, 0.75, 0.975), digits_summary=2):
         if fit.mode == 1:
             return "Stan model '{}' is of mode 'test_grad';\n"\
                    "sampling is not conducted.".format(fit.model_name)
@@ -76,6 +76,11 @@ def _print_stanfit(fit, pars=None, probs=(0.025, 0.25, 0.5, 0.75, 0.975), digits
                                s['summary_colnames'], digits_summary)
         return header + body + footer
 
+def _print_stanfit(fit, pars=None, probs=(0.025, 0.25, 0.5, 0.75, 0.975), digits_summary=2):
+    # warning added in PyStan 2.17.0
+    warnings.warn('Function `_print_stanfit` is deprecated and will be removed in a future version. '\
+                  'Use `stansummary` instead.', DeprecationWarning)
+    return stansummary(fit, pars=None, probs=(0.025, 0.25, 0.5, 0.75, 0.975), digits_summary=2)
 
 def _array_to_table(arr, rownames, colnames, n_digits):
     """Print an array with row and column names
--- a/pystan/misc.py
+++ b/pystan/misc.py
@@ -80,7 +80,7 @@ def _print_stanfit(fit, pars=None, probs=(0.025, 0.25, 0.5, 0.75, 0.975), digits
     # warning added in PyStan 2.17.0
     warnings.warn('Function `_print_stanfit` is deprecated and will be removed in a future version. '\
                   'Use `stansummary` instead.', DeprecationWarning)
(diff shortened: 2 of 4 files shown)


ahartikainen:
Do we want to add test for this? `Under test_misc_args.py`
e.g.

    pystan.misc._print_stanfit() != pystan.misc._print_stanfit(pars='alpha')

ariddell:
Looks good! A test would be great.

Maybe expose a function (``fit.summary(..)``) which makes this easier to call? (R uses ``summary(fit)``, right?) We should probably add some docs so people can find this somewhere. I can't recall how difficult it is to get Sphinx to look into the docstrings in a Cython file (which is where we'd have to put the summary method). This could wait though.

ahartikainen:
I added the test. In the test it uses `fit.summary()` / `fit.summary(pars=...)`

Problem is that `fit.summary` is currently used and returns data. Do we want to break this and return string instead or create a keyword to decide what return type is wanted (default str). Then we can implement `pandas dataframe` or html printing for notebooks if needed.

Also do we want to print the string or just to return it?

    # currently returns this line
    pystan.misc._summary(self, pars, probs)

Pandas dataframe for normal table and table for each chain

    import pandas as pd
    summary_dict = pystan.misc._summary(self, pars, probs)
    df = pd.DataFrame(data=summary_dict['summary'], 
                           index=summary_dict['summary_rownames'], 
                           columns=summary_dict['summary_colnames'])

    chain_summary_list = []
    for chain in range(summary_dict['c_summary'].shape[-1]):
        table_chain = pd.DataFrame(data=summary_dict['c_summary'][:, :, chain],
                                        # rename of the variable can be omitted
                                        index=["{}_chain_{}".format(name, chain) for name in summary_dict['c_summary_rownames']],
                                        columns=summary_dict['c_summary_colnames'])
        chain_summary_list.append(table_chain)

    return df, chain_summary_list

ariddell:
Let's follow the principle of least surprise and keep fit.summary mirroring RStan's [summary method](https://github.com/stan-dev/rstan/blob/6426e7a3f763bf6f18fedc8d91e097287ef1ef16/rstan/rstan/man/stanfit-method-summary.Rd).

Let's turn the "private" pystan.misc._print_stanfit(fit) into a public function that people can use. Should we call it ``pystan.misc.print_summary`` or ``pystan.misc.stansummary`` (the cmdstan binary is called ``stansummary``)?

If we change its name we'll need to have the fit objects use the new version. The lines to change are these in the pyx template file (stanfit4model):

```
    def __unicode__(self):
        # for Python 2.x
        return pystan.misc._print_stanfit(self)

    def __str__(self):
        s = pystan.misc._print_stanfit(self)
        return s.encode('utf-8') if PY2 else s
```

ahartikainen:
I used `stansummary`. In my opinion `print_summary` should print it, but now `stansummary` will return a string. Are we going to only keep it in `.misc` or import it to main namespace?

Do we want to have a html table option for jupyter notebook?

Where in the docs a mention of this function should go?

ariddell:
Here's something I sent via email but which Github seems to have ignored:

```
If you give it a good docstring Sphinx can pick it up. Look at model.py
for the docstring format. The docstring gets rendered if you point
Sphinx at the function (or method or class) (in doc/api.rst).

(Things don't work quite as smoothly as they should because Sphinx
doesn't know how to read the docstrings inside of .pyx files.)
```

Putting stansummary in the main namespace sounds like a good idea. There's no reason it needs to live in ``pystan..misc``.

ariddell:
This looks great. I feel like we should add docs for this particular function since it's in the main namespace. Does that make sense?

ahartikainen:
Yes, I agree. I will do the docstring tomorrow.

This function returns a string. Is it confusing if we are adding it to main namespace?

    print(pystan.stansummary(fit)) vs pystan.stansummary(fit)

also could add method to fit object?

    print(fit.stansummary(pars=...))

ariddell:
I'm leaning towards returning a string. I think this is what everyone else does when they have summary output. For example, statsmodels has ``print(results.summary())`` (from http://www.statsmodels.org/stable/)

ahartikainen:
I added the docstring to `pystan.misc.stansummary` and also updated the `StanFit4Model` (`fit.stansummary()`). 

The method for `fit` object is for convenience: if one does this: `from pystan import StanModel`. 

We should update our printed parameters to match closer the `CmdStan` output, but it should be done later in another PR.

ahartikainen:
What is the problem with travis. It will not run all our test anymore (timeout)? Are their servers too full?

Not sure if there are any problems with this PR. It should be merged before next release (2.17)

ariddell:
Let me see if there's an easy way to split up the tests so travis doesn't time out.

ahartikainen:
Everything looks ok. Can I merge?

PR #369, opened 2017-09-08
CI: split tests up, avoid CI timeout

ariddell:
Tests need to finish in less than 50 minutes. This commit splits the tests up into two batches each of which should finish in less than 50 minutes.

``nosetests`` is used in the tests currently. In the future, some more modern test framework should be used.

Before merging, check to see if any tests are getting lost. The total number of tests should be 95.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,14 @@ compiler:
   - clang
   - gcc
 env:
-  - TRAVIS_PYTHON_VERSION=2.7
-  - TRAVIS_PYTHON_VERSION=3.3
-  - TRAVIS_PYTHON_VERSION=3.4
-  - TRAVIS_PYTHON_VERSION=3.5
-  - TRAVIS_PYTHON_VERSION=3.6
+  - TRAVIS_PYTHON_VERSION=2.7 TESTS_EXCLUDE_RE='^test_[a-p].*\.py'
+  - TRAVIS_PYTHON_VERSION=3.4 TESTS_EXCLUDE_RE='^test_[a-p].*\.py'
+  - TRAVIS_PYTHON_VERSION=3.5 TESTS_EXCLUDE_RE='^test_[a-p].*\.py'
+  - TRAVIS_PYTHON_VERSION=3.6 TESTS_EXCLUDE_RE='^test_[a-p].*\.py'
+  - TRAVIS_PYTHON_VERSION=2.7 TESTS_EXCLUDE_RE='^test_[^a-p].*\.py'
+  - TRAVIS_PYTHON_VERSION=3.4 TESTS_EXCLUDE_RE='^test_[^a-p].*\.py'
+  - TRAVIS_PYTHON_VERSION=3.5 TESTS_EXCLUDE_RE='^test_[^a-p].*\.py'
+  - TRAVIS_PYTHON_VERSION=3.6 TESTS_EXCLUDE_RE='^test_[^a-p].*\.py'
 notifications:
   email:
     recipients:
@@ -25,4 +28,4 @@ notifications:
     on_success: change
     on_failure: always
 install: source continuous_integration/install.sh
-script: bash continuous_integration/test_script.sh
+script: bash continuous_integration/test_script.sh --exclude="$TESTS_EXCLUDE_RE"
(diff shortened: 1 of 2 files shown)


ahartikainen:
Looks good.

ariddell:
I think several tests aren't getting run due to some problem with
nosetests. I'm tempted to say that having most tests run is better than
hitting the timeouts.

ahartikainen:
What tests are not running?

I counted 26 + 50 = 76

ariddell:
Looking at a commit from a version before this PR, it looks like there
are 94 tests in total. for example,
https://travis-ci.org/stan-dev/pystan/jobs/246289982

Should we just merge this so we get working tests?

ahartikainen:
Yes, we can fix it later.

ahartikainen:
Hi,

we could add this to `continuous_integration/test_script.sh`, before any tests are really run

    nosetests -v --collect-only $@

It will output the name of the tests and the total number.

Takes ~1min to run for 94 tests.

ahartikainen:
Hi, 

I found atleast one error.

There are tests called `test_8school...`. So to fix this change

    [a-p] --> [a-p0-9]

The other thing is, that we have 89 tests. The nose is printing some info (e.g. `'Make sure labels are not getting switched ... ok'`) and counts those as a test.

The other error is probably coming from the way `--exclude=..` works. At the same time, it checks the filename and the test function name.

These are the test that are missed if `[a-p0-9]` is used

    'test_grad_log (test_rstan_stanfit.TestStanfit) ... ok',
    'test_init_zero_exception_inf_grad (test_rstan_stanfit.TestStanfit) ... ok',
    'test_optim_stdnorm (test_rstan_testoptim.TestOptim) ... ok',
    'test_optim_stdnorm_bfgs (test_rstan_testoptim.TestOptim) ... ok',
    'test_optim_stdnorm_from_file (test_rstan_testoptim.TestOptim) ... ok',
    'test_optim_stdnorm_lbfgs (test_rstan_testoptim.TestOptim) ... ok',
    'test_remove_empty_pars (test_misc.TestMisc) ... ok',
    'test_rhat (test_chains.TestChains) ... ok',
    'test_rhat (test_chains.TestChainsNormal) ... ok',
    'test_rhat_zero (test_chains.TestChains) ... ok',
    'test_rstan_read_rdump (test_misc.TestMisc) ... ok',
    'test_stan_rdump (test_misc.TestMisc) ... ok',
    'test_stan_rdump_array (test_misc.TestMisc) ... ok'

So `test_grad_log` is in `[^a-p0-9]` but `test_rstan_stanfit` is not.

Not sure how to go forward with this? Can we somehow fix the regex? Should we rename our tests?

ariddell:
Thanks. I did try the ``--collect`` parameter at one point. Perhaps this
could be used to figure out what tests are being excluded.

ariddell:
I'll add the 0-9. Maybe renaming the tests would be the easiest in the
short term.

In the medium-term we need to move away from nose because nose is no
longer being developed. I've been using py.test in httpstan and the
pystan 3 prototype. It's working well.

ariddell:
Now we've got 26 + 64 = 80 tests being run now. Let's just rename the tests. I'll try that now.

ahartikainen:
Hi, these should fix the problem with `--exclude`. This will first ignore the files not used and then include that are (also the default ignore has been added)

Run test for filenames `^test_[a-p]`
    --ignore="^test_[^a-p0-9]" --ignore="^\\." --ignore="^_" --ignore="^setup\\.py$" --include="^test_[a-p0-9]"

Run test for filenames `^test_[q-z]`
    --ignore="^test_[a-p0-9]" --ignore="^\\." --ignore="^_" --ignore="^setup\\.py$" --include="^test_[^a-p0-9]"

ahartikainen:
Weird. Now it runs all tests?

    Ran 94 tests in 1651.347s

Tested this on my pc (mac) and it did filter correctly.

ariddell:
Maybe we need a specific version of nose. I have at least one version of
nosetests which has an --ignore-files argument but no --ignore argument.

ariddell:
This is really frustrating. I'll rename the tests today.

ariddell:
I think I fixed it. Let's see if it works.

Issue #364, opened 2017-07-07
docs: Update Windows installation / support


ahartikainen:
Update the installation notes to use conda distribution with MSYS2 GCC compiler (`m2w64 toolchain`). Add disclaimer/warning concerning MSVC + bugs.

We should aim for something similar that theano uses: http://deeplearning.net/software/theano/install_windows.html

- Anaconda / Miniconda installer
    -  example with conda environment
- GCC step: `conda install -c msys2 m2w64-toolchain=5.3.0` to install MSYS2 GCC
- Other packages either in the first step or with explicit example with `conda-forge`
- TODO: Do we suggest to use `distutils.cfg` (this can leak to other environments) or some flag in source code? `CC='gcc'`?

Examples

    # Optional - create a conda environment
    conda create -n stan_env python=3.6
    source activate stan_env
    
    #dependencies
    conda install numpy cython matplotlib scipy pandas
    
    # gcc compiler components
    conda install libpython
    conda install -c msys2 m2w64-toolchain=5.3.0
    
    # step with distutils.cfg
    ... (can we somehow automate this?)
    
    # pip / conda
    pip install pystan

~~We could also try to do conda step with pip.~~

~~pip install conda~~
~~conda install libpython~~
~~conda install -c msys2 m2w64-toolchain=5.3.0~~

ariddell:
If you've verified this works, we should certainly add it now. Thanks for digging into this.

lwiklendt:
Fantastic. Looks like n_jobs > 1 works too.

ahartikainen:
We have to either remove the lines 286-287 in `Model.py` or update the if statement to recognize if the compiler is `msvc` or `gcc`.

    if platform.platform().startswith('Win'):
        extra_compile_args = ['/EHsc', '-DBOOST_DATE_TIME_NO_LIB']

Do we have any reliable way to do it before compilation or can we add a keyword if msvc compiler is used?

Also at the same time we could update line 29 so it would first try to import Extension from `setuptools` and if that fails then from `distutils`. This would fix the issue with msvc 14.1 / 2017 if user has updated the `setuptools`.

In my opinion we should not break the msvc compilation but only give warning?

dimitry12:
For me the instructions didn't work when using 64-bit Anaconda 4.2.2 Python 3.5 on Windows:

- after GCC compilation completed,
- I was getting `TypeError: Shared Cython type generator is not a type object` at:
[...]

PR #371, opened 2017-09-13
Add Windows gcc support

ahartikainen:
Add msvc 2017 support by importing from `setuptools`. Needs `setuptools >= 34.4.0`.

Add compiler keyword (default None). User can choose mingw32, gcc / msvc.

On windows, check with subprocess (quietly) if gcc is installed on the system. If found, use it else use 'msvc'.

#### Summary:

Update model.py to support gcc on Windows.
Also update for msvc 2017 (setuptools).

See #364
#### Intended Effect:

Support gcc (MSYS2)

#### How to Verify:

Run on Windows with supported gcc compiler (MSYS2)

#### Side Effects:

Automatically detect if gcc installed with `subprocess`

#### Documentation:

Needs to be updated.

#### Copyright and Licensing

### DIFF
--- a/pystan/model.py
+++ b/pystan/model.py
@@ -26,11 +26,19 @@ import time
 import warnings
 
 import distutils
-from distutils.core import Extension
+
+try:
+    # MSVC 2017 support (setuptools >= 34.4.0)
+    from setuptools import Extension
+except ImportError:
+    from distutils.core import Extension
 
 import Cython
 from Cython.Build.Inline import _get_build_extension
 from Cython.Build.Dependencies import cythonize
+
+# GCC detection
+import subprocess
 
 import numpy as np
 
@@ -134,7 +142,12 @@ class StanModel:
     verbose : boolean, False by default
         Indicates whether intermediate output should be piped to the console.
         This output may be useful for debugging.
-
+    
+    compiler : string, None by default
+        Only for Windows. Choose compiler type. 
+        - for GCC use 'gcc' or 'mingw32', 
+        - for Microsoft Visual C++ use 'msvc'
+    
     kwargs : keyword arguments
         Additional arguments passed to `stanc`.
 
@@ -202,7 +215,7 @@ class StanModel:
     def __init__(self, file=None, charset='utf-8', model_name="anon_model",
                  model_code=None, stanc_ret=None, boost_lib=None,
                  eigen_lib=None, verbose=False, obfuscate_model_name=True,
-                 extra_compile_args=None):
+                 extra_compile_args=None, compiler=None):
 
         if stanc_ret is None:
             stanc_ret = pystan.api.stanc(file=file,
@@ -274,6 +287,10 @@ class StanModel:
             ('BOOST_NO_DECLTYPE', None),
             ('BOOST_DISABLE_ASSERTS', None),
         ]
+        
+        # create here to allow compiler injection on Windows
+        build_extension = _get_build_extension()
+        
         # compile stan models with optimization (-O2)
         # (stanc is compiled without optimization (-O0) currently, see #33)
         if extra_compile_args is None:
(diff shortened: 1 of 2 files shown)


ahartikainen:
One problem that needs to be solved is the docs. Do we edit the current one, or keep it as is and create new instructions to beginning? This also needs disclaimer for Visual Studio + Eigen incompatibility.

Other problem is that I'm not sure will this affect the `cygwin` distributions.

Also we probably need to update the appveyor and add some tests.

ariddell:
You've got a better understanding of things than I do.

I think we should focus on getting one compiler to work reliably and then worry about other compilers/distributions.

On this patch specifically, I think the canonical way of setting the compiler is to use the CC environment variable (see, for example, https://stackoverflow.com/questions/16737260/how-to-tell-distutils-to-use-gcc ). Perhaps we could have windows users do that instead of specifying the compiler as a string?

ariddell:
I wish we knew how many people are using the wheels vs. compiling from scratch. If everyone has moved over to the wheels I would suggest adding setuptools>=34.4 to the requirements.

ahartikainen:
If I remember correctly, the `CC` env var does not work on windows. The better way for forcing `mingw32` could be `distutils.cfg` file where one would put this

    [compiler]
    mingw32

But, this is on Windows. And finding/creating that file to the correct distutils folder (also I have noticed that this can pollute other environments).

We could force the user input `compiler=mingw32`/ `compiler=msvc` or let cython to decide. Then no subprocess command is needed.

ahartikainen:
The setuptools is not really a problem. It works even with older version. Only msvc 2017 needs the updated one.

ariddell:
What about giving very specific instructions on how to edit distutils.cfg? I'm a little wary of stepping on setuptools' toes by overriding things. But I'll trust your judgment here. If this is the easiest way to do it, let's do it.

I think fully supporting one compiler is the priority. If we can support two, of course, that's great.

ahartikainen:
I guess we could create the very specific instructions for distutils.cfg. Then no need for the "injection".

Then we could call `distutils.ccompiler.get_default_compiler()` for default compiler. This is needed for `extra_compile_args` (msvc vs gcc)

They (`python`) have recently added support for msvc 2017 for distutils, but not sure when they are going to update it. 

https://github.com/python/cpython/tree/master/Lib/distutils

The problem with compilers are: 
regular Pystan user should use `gcc` for vectorized functions with conda on Windows. 
Other software with manually tuned models (fbprophet) bypass the problem. So if we break the support for msvc then we probably are going to break their software. 

My guess is that not everybody are using conda on Windows. If this is true, then they can only use MSVC.

ariddell:
It looks like MSVC 2017 support will be in the next point release of
Python 3.5 and 3.6 based on https://github.com/python/cpython/pull/1632

I'm a little worried about trying to support gcc (and answer support
problems) before we have MSVC working. If you're not bothered by this,
however, we can merge this and update the documentation.

ariddell:
Looks good. Should we add those msvc ``if`` statements to setup.py as well?

ahartikainen:
Then we can remove the `setuptools` part and wait for the next release. (I will do this now).

So this PR will only enable the gcc (testing if default compiler is `msvc` (default) or mingw32 (`distutils.cfg`). We can update the rest of the docs etc in another PR.

One problem is, that `distutils.ccompiler` does not work as I assumed. So I will go back to `extension.compiler` method. That will work when we don't need any conditional imports.

Yes, we kind of need a way to do this in `setup.py`. One way is to do this is the same way we are doing it in the `model.py`.

    # add this inside the if windows statement?
    from Cython.Build.Inline import _get_build_extension
    compiler = _get_build_extension().compiler

Can we import `Cython` in the `setup.py` file?

ariddell:
In general, it looks good to me.

Before merging, I'd recommend making sure that appveyor is working (or is failing as expected).

ariddell:
Possibly of interest:

- Appendix: Installing MinGW on Windows http://docs.cython.org/en/latest/src/tutorial/appendix.html
- cython-users discussion: https://groups.google.com/forum/#!searchin/cython-users/Cython$20with$20MinGW-w64%7Csort:relevance/cython-users/39ibHNjFr_g/61cbmUQdAgAJ

ahartikainen:
Ok, everything looks great. I will merge this. Let's update the docs in another PR.

Currently, (with default) everything works with `msvc` as before. So this PR only enables `gcc` (`mingw32` for python <3.4 / 2.7 and `MSYS2-mingw64` for python 3.5, 3.6). User needs to create `distutils.cfg` either to `home` directory or `distutils` folder (`MSYS2-mingw64` is possible creating this automatically). I will test what different options will work before updating the docs.

It would be great if there was some extra command to get `msvc` to compile `Eigen` stuff.

PR #377, opened 2017-09-22
[DOC] Update windows instructions

ahartikainen:
Add warning for MSVC support (vectorized functions)

Instructions for GCC on Windows

#### Documentation:

Update only the docs

#### Copyright and Licensing

### DIFF
--- a/doc/windows.rst
+++ b/doc/windows.rst
@@ -81,7 +81,8 @@ Supported GCC compilers
 
 For mingw32 installation see
 http://docs.cython.org/en/latest/src/tutorial/appendix.html
-or with conda use
+
+or with conda use::
 
     conda install mingw32 libpython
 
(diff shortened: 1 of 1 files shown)


ahartikainen:
@ariddell, please edit as you see fit.

I added instructions for mingw32 (Python 2.7, 3.4). Not sure if they are needed.

ariddell:
Perhaps we should mention what's the problem with vectorized operations and when it might be fixed? (If it's never going to be fixed, we should perhaps drop support for Windows, right?)

ahartikainen:
We could do that.

Maybe drop the official support? Some applications do work on Windows (fbprophet).

Maybe we could test which version of the msvc is not working.

ahartikainen:
If we are going to drop windows support, let's drop the support in the later release.

ariddell:
Is there a Boost bug open about this? Or a Microsoft bug issue? We could say something like "until this bug is fixed" PyStan is not officially supported on Windows. We can still compile and distribute binary wheels, of course.

ahartikainen:
Ouch, it's Eigen problem, not Boost.

I guess it's Microsoft bug / Stan math codestyle problem. Not totally sure who to really "blame".

ahartikainen:
I added "partially supported". Because it's working for a subset of functions just fine.

Should we change it unsupported?

Not sure if there is a bug issue open.

ariddell:
I think that we do not want to encourage people to use PyStan on Windows until the bug is fixed. Partial support or unsupported is fine by me.

I think the decision is up to you. If you don't mind answering Windows users' questions on Github/Discourse then "partial support" sounds about right.

ahartikainen:
I don't mind that. 

And also, one can get full set of functions on Windows, if they install the M2W64 toolchain.

Also, when PyStan3 comes, could we force people to use the gcc (this is what RStan does) by either installing the conda version or the RTools version.

In my opinion, this PR can be merged after the test have been completed.

PR #381, opened 2017-09-29
[WIP] Add DeprecationWarning for pystan.stan

### LINKED ISSUE #327: Add DeprecationWarning for stan function
#### Summary:

Users should first construct an instance of ``StanModel`` and then call ``sampling`` or ``optimizing``. Use of the ``stan`` function should be discouraged.
- [ ] Add DeprecationWarning
- [ ] Update docs

#### Description:
Use of ``stan`` alone is undesirable for many reasons. It makes it difficult to save/cache model compilation, for example. A separate construction of a ``StanModel`` instance is more in keeping with Python conventions (see, for example, scikit-learn).

ariddell:
Closes #327.

TODO:

- [ ] Update docs. Remove any use of `pystan.stan`.


ariddell:
The DeprecationWarning has been added and the docs are updated. DeprecationWarnings are not shown by default (new information for me). One has to run ``python -W script.py`` to see them. (But I believe some testing frameworks turn on display of warnings by default.)

ariddell:
There's an odd error with Python 2.7 and numpy. It looks like it might be matplotlib related.

Issue #385, opened 2017-10-06
New minimum version for numpy needed?


ariddell:
CI is giving this error under Python 2.7 only:

```
  File "/home/travis/miniconda3/envs/env_name/lib/python2.7/site-packages/pystan/plots.py", line 15, in traceplot

    from pystan.external.pymc import plots

  File "/home/travis/miniconda3/envs/env_name/lib/python2.7/site-packages/pystan/external/pymc/plots.py", line 13, in <module>

    from pylab import *

  File "/home/travis/miniconda3/envs/env_name/lib/python2.7/site-packages/pylab.py", line 1, in <module>

    from matplotlib.pylab import *

  File "/home/travis/miniconda3/envs/env_name/lib/python2.7/site-packages/matplotlib/pylab.py", line 230, in <module>

    import matplotlib.finance

  File "/home/travis/miniconda3/envs/env_name/lib/python2.7/site-packages/matplotlib/finance.py", line 24, in <module>

    from matplotlib.dates import date2num

  File "/home/travis/miniconda3/envs/env_name/lib/python2.7/site-packages/matplotlib/dates.py", line 136, in <module>

    import matplotlib.ticker as ticker

  File "/home/travis/miniconda3/envs/env_name/lib/python2.7/site-packages/matplotlib/ticker.py", line 181, in <module>

    from matplotlib import transforms as mtransforms

  File "/home/travis/miniconda3/envs/env_name/lib/python2.7/site-packages/matplotlib/transforms.py", line 39, in <module>

    from matplotlib._path import (affine_transform, count_bboxes_overlapping_bbox,

ImportError: numpy.core.multiarray failed to import
```
The CI is using numpy 1.7.1 in this case, which is an old version of numpy.

Perhaps PyStan could use a newer version of numpy. The version used in Ubuntu 16.04 or Debian 8 might be a good reference point.

ariddell:
Debian 8 uses 1.8.2 (Source: https://packages.debian.org/jessie/python-numpy )

PR #387, opened 2017-10-06
CI: update numpy version to 1.8.2 for Python 2 on CI.

ariddell:
With luck fixes issue described in #385 .


ahartikainen:
Should we change the `requirements.txt`, or is this just a problem with `numpy` + `matplotlib` interaction?

ariddell:
Let's warn about the updated numpy requirement in 2.18 and then update
it in the requirements.txt for 2.19? I don't want to break pystan for
someone running it using an old numpy (on, say, a linux cluster running
RHEL or Centos)

ahartikainen:
Sounds good.

PR #425, opened 2018-02-05
Output samples to pandas dataframe

braaannigan:
#### Summary:
Samples can be output as a pandas dataframe.  This can be done as a method on a fit object or equivalently as a function that takes fit as an argument.  

#### Intended Effect:
The samples are output as a dataframe with the sample number as the index.

#### How to Verify:
Once the model is fit (e.g. 8 schools) can run
```
df = pystan.misc.to_dataframe(fit)
```
or more simply
```
df = fit.to_dataframe()
```
Tests to follow once reviewer suggestions have been taken onboard.

#### Side Effects:
The dependence on pandas is through a try/except/raise statement.  If pandas isn't found it prints an error message explaining how to install pandas.  In some versions of the **iPython** shell the ImportError statement raises its own error with a message like:
```
AttributeError: 'ImportError' object has no attribute '_render_traceback_'
```
I think this is the same as this error flagged with ipython:https://github.com/ipython/ipython/issues/9978

If the command is run more than once then the error message appears as it was intended.  Hopefully, this will be be resolved in future versions of ipython.

#### Documentation:
To follow once reviewer suggestions have been taken onboard.

#### Reviewer Suggestions: 

### DIFF
--- a/pystan/tests/test_extract.py
+++ b/pystan/tests/test_extract.py
@@ -134,38 +134,34 @@ class TestExtract(unittest.TestCase):
         for idx in range(2):
             for jdx in range(3):
                 name = 'alpha_{}_{}'.format(idx+1,jdx+1)
-                assert_series_equal(df[name],pd.Series(alpha[:,idx,jdx],name=name))
+                assert_array_equal(df[name].values,alpha[:,idx,jdx])
         for idx in range(2):
             name = 'beta_{}'.format(idx+1)
-            assert_series_equal(df[name],pd.Series(beta[:,idx],name=name))
-        assert_series_equal(df['lp__'],pd.Series(lp__,name='lp__'))
-        ss = self.fit.extract(permuted=True)
-        alpha = ss['alpha']
-        beta = ss['beta']
-        lp__ = ss['lp__']
+            assert_array_equal(df[name].values,beta[:,idx])
+        assert_array_equal(df['lp__'].values,lp__)
         # Test pars argument
         df = self.fit.to_dataframe(pars='alpha')
         self.assertEqual(df.shape, (4000,6))
         for idx in range(2):
             for jdx in range(3):
-                name = 'alpha_{}_{}'.format(idx,jdx)
-                assert_series_equal(df[name],pd.Series(alpha[:,idx,jdx],name=name))
+                name = 'alpha_{}_{}'.format(idx+1,jdx+1)
+                assert_array_equal(df[name].values,alpha[:,idx,jdx])
         # Test pars and dtype argument
         df = self.fit.to_dataframe(pars='alpha',dtypes = {'alpha':np.int})
         alpha_int = ss['alpha'].astype(np.int)
         self.assertEqual(df.shape, (4000,6))
         for idx in range(2):
             for jdx in range(3):
-                name = 'alpha_{}_{}'.format(idx,jdx)
-                assert_series_equal(df[name],pd.Series(alpha_int[:,idx,jdx],name=name))
+                name = 'alpha_{}_{}'.format(idx+1,jdx+1)
+                assert_array_equal(df[name].values,alpha_int[:,idx,jdx])
 
     def test_to_dataframe_permuted_false_inc_warmup_false(self):
         fit = self.fit
         ss = fit.extract(permuted=False)
-        df = fit.to_dataframe()
+        df = fit.to_dataframe(permuted=False)
         num_samples = fit.sim['iter'] - fit.sim['warmup']
         num_chains = fit.sim['chains']
-        self.assertEqual(df.shape, (num_samples*num_chains,9))
+        self.assertEqual(df.shape, (num_samples*num_chains,17))
         alpha_index = 0
         for jdx in range(3):
             for idx in range(2):
@@ -183,35 +179,30 @@ class TestExtract(unittest.TestCase):
                 )
             for n in range(1,num_chains+1):
                 assert_array_equal(df.loc[df.chain == n,'lp__'].values,ss[:,n-1,-1])
-            for n in range(num_chains):
-                assert_array_equal(
+        diagnostic_type = {'divergent':bool,'energy':float,'treedepth':int,
+			                'accept_stat':float, 'stepsize':float, 'n_leapfrog':int}
+        for n in range(num_chains):
+            assert_array_equal(
                 df.chain.values[n*num_samples:(n+1)*num_samples],
                 (n+1)*np.ones(num_samples,dtype=np.int)
                 )
-                assert_array_equal(
+            assert_array_equal(
                 df.chain_idx.values[n*num_samples:(n+1)*num_samples],
                 np.arange(1,num_samples +1,dtype=np.int)
                 )
+            for diag, diag_type in diagnostic_type.items():
                 assert_array_equal(
-                df.divergent__.values[n*num_samples:(n+1)*num_samples],
-                fit.get_sampler_params()[n]['divergent__'][-num_samples:].astype(bool)
-                )
-                assert_array_equal(
-                df.energy__.values[n*num_samples:(n+1)*num_samples],
-                fit.get_sampler_params()[n]['energy__'][-num_samples:]
-                )
-                assert_array_equal(
-                df.treedepth__.values[n*num_samples:(n+1)*num_samples],
-                fit.get_sampler_params()[n]['treedepth__'][-num_samples:].astype(int)
+                df[diag+'__'].values[n*num_samples:(n+1)*num_samples],
+                fit.get_sampler_params()[n][diag+'__'][-num_samples:].astype(diag_type)
                 )
 
     def test_to_dataframe_permuted_false_inc_warmup_true(self):
         fit = self.fit
         ss = fit.extract(permuted=False, inc_warmup = True)
-        df = fit.to_dataframe()
+        df = fit.to_dataframe(permuted=False,inc_warmup=True)
         num_samples = fit.sim['iter']
         num_chains = fit.sim['chains']
-        self.assertEqual(df.shape, (num_samples*num_chains,9))
+        self.assertEqual(df.shape, (num_samples*num_chains,17))
         alpha_index = 0
         for jdx in range(3):
             for idx in range(2):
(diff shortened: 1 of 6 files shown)


braaannigan:
Looking at the tests, I see that parameters can be more than two dimensional.  Would it be better to direct those cases to the forthcoming to_xarray method?

ahartikainen:
I think column naming convention could follow `fit.flatnames`? (permuted=False).

So it is fine to ravel nD arrays to table.

We could have one option to also extract diagnostig info; divergent, chain, chain_idx lp__, warmup etc. Naming could follow lp__ with two __ at the end of the name.

braaannigan:
Thanks @ahartikainen, I'll submit an updated version in the next day or 2

ahartikainen:
Hi,

We could have also one option for permuted=False. I recently merged PR (it was a hybrib between permuted=True and permuted=False) so your code could follow that one (for option permuted=False). The end result is (samples, chains, pars) so you would still need unravel the parameters. Reshape should be able to do it. This also makes easier to implement the chain_id and chain_idx values if needed for diagnostic.

Should we make a list of all the diagnostic stuff we want to add to the frame? (cc.  @ariddell @betanalpha @seantalts )

 - chain_id # int
 - chain_id_idx # int << original index for each chain
 - divergent # bool
 - energy
 - lp__
 - treedepth # int

After this PR we can make another one for .summary_to_dataframe which could either wrap the current summary method or even make a new one with added information to (updated to current CmdStan)

braaannigan:
Hi @ahartikainen 

permuted=False seems like it would be straightforward to add.  Any suggestions for the naming convention? Not sure that there's anything snappier than 'alpha_0_0_chain_0'.

For the diagnostics, are these the parameters in fit.get_sampler_params()?

For chain_id and chain_id_idx, I haven't seen these saved anywhere, would they have to be constructed when the dataframe is made?

Liam

braaannigan:
Also - presumably we'll need the same inc_warmup argument for the permuted = False case

ahartikainen:
I was thinking that we could put all chains in one column and then create one new column where it would say which chain it is.

I think that `theta_0_0` is good. I think we should follow the Stan language and start the indexes at 1 --> so `theta_1_1`.

So for naming:

Scalar values: `y`
Vector/1D arrays: `y_1, y_2`
Matrices, 2D arrays: `y_1_1`, `y_1_2` ...
nD arrays: `y_1_1_1` etc. 

Yes, chain idx is just the original index (permuted=False)
 
e.g. `df`

    chain    chain_idx    y    divergent__
    1           1        0.3    False
    1           2        0.4    False
    1           3        0.2    False
    2           1        0.2    False
    2           2        0.2    False
    2           3        0.3    False
    3           1        0.4    False

and this way one can use `df.groupby(by='chain').mean()` etc

ariddell:
The chain_id is assigned/used by Stan, so we have to use that. It's in
the ``sim`` object/dict, IIRC.

ariddell:
We might want to consider aligning with what RStan does when it outputs
a data.frame. I suspect they do use theta.1.1.

Did someone mention using a pandas.Panel instead of a DataFrame? It
makes sense to me to put the chains on a different axis.

pystan/tests/test_extract.py
```diff
@@ -96,3 +110,49 @@ def test_extract_dtype(self):
...
+        pars = ['alpha', 'beta', 'lp__']
+        ss = self.fit.extract(pars=pars, dtypes = dtypes, permuted=False)
+        alpha = ss['alpha']
+        beta = ss['beta']
+        lp__ = ss['lp__']
+        self.assertEqual(alpha.dtype, np.dtype(np.int))
+        self.assertEqual(beta.dtype, np.dtype(np.int))
+        self.assertEqual(lp__.dtype, np.dtype(np.float))
+
+  def test_to_dataframe(self):
```
ahartikainen: Indentation error here

braaannigan:
Hi

I've added the diagnostics in the permute=false branch. 

There is one issue that is stopping me from running the tests.  When I create the dataframe using 
```
df = pystan.misc.to_dataframe(fit,...)
```
it works properly with any combination of arguments.  However, when I call it the way that most people will use it as a method on ```fit```, then the default arguments from stanfit4model.pyx aren't passed through properly.    The default arguments are:
```
pars=None, permuted=True, dtypes=None, inc_warmup=False
```
When I call fit.to_dataframe(), and print out the arguments using a print statement in pystan.misc.to_dataframe they are 
```
pars=None, permuted=None, dtypes=None, inc_warmup=False
```
So ```permuted``` is coming through as ```None``` instead of ```True``` while ```inc_warmup``` does come through as a boolean.  Any suggestions as to how to what's going on?

ahartikainen:
Hi, code looks great. Here are some comments:

maybe the True --> None is a cython bug?

Could we add diagnostic keyword, default could be true.

Also option for pars + permuted=False.

braaannigan:
Have a bit of travelling at the mo, will take a few days to sort this out.  Still welcoming suggestions about the cython issue though!

ahartikainen:
I can try to figure out the Cython issue.

braaannigan:
I've found the cause of the cython issue: I was doing my quick test using a model that had been compiled and pickled before I made some changes, so subsequent changes weren't showing up.  I'm trying to get the rest of the points we discussed together for a new commit.

ahartikainen:
Good that is fixed. I have done that mistake multiple times too.

braaannigan:
For the ```fit.extract()``` method the ```pars``` option is described as 
```
pars : {str, sequence of str}
           parameter (or quantile) name(s)
```
Is this correct that you can pass quantiles as the argument e.g. something like ```pars = [25,75]```?

ahartikainen:
I don't believe quantiles work. But pars could contain keys that are not parameters: e.g. `lp__`

Diagnostics could also get 

    stepsize__
    n_leapfrog__

braaannigan:
Can we go to merge at this point?

ahartikainen:
Everything else seems ok, still need to update `api.rst` --> StanFit4model section (docstring for to_dataframe).

braaannigan:
Ok, will probably be tomorrow before I do that.  

I'll be following up with some example documentation in the next few weeks in a separate PR as well.

ahartikainen:
Sounds great!

ahartikainen:
Hi, tested this in a new conda environment.

    import pystan
    model_code ="""
    data {
      int<lower=0> J;
      real y[J];
      real<lower=0> sigma[J];
    }

    parameters {
      real mu;
      real<lower=0> tau;
      real theta[J];
    }

    model {
      mu ~ normal(0, 5);
      tau ~ cauchy(0, 5);
      theta ~ normal(mu, tau);
      y ~ normal(theta, sigma);
    }

    """

    stan_model = pystan.StanModel(model_code=model_code)
    schools_dat = {'J': 8,
               'y': [28,  8, -3,  7, -1,  1, 18, 12],
               'sigma': [15, 10, 16, 11,  9, 11, 10, 18]}
    fit = stan_model.sampling(data = schools_dat)

then I did this

    df = fit.to_dataframe()

or this

    df = fit.to_dataframe(permuted=True)

and it raises Exception.  `permuted=False` works as expected.

Error message here:

    ---------------------------------------------------------------------------
    Exception                                 Traceback (most recent call last)
    ~/miniconda3/envs/stantmp2/lib/python3.6/site-packages/pandas/core/frame.py in _ensure_valid_index(self, value)
       2563             try:
    -> 2564                 value = Series(value)
       2565             except:

    ~/miniconda3/envs/stantmp2/lib/python3.6/site-packages/pandas/core/series.py in __init__(self, data, index, dtype, name, copy, fastpath)
        263                 data = _sanitize_array(data, index, dtype, copy,
    --> 264                                        raise_cast_failure=True)
        265 

    ~/miniconda3/envs/stantmp2/lib/python3.6/site-packages/pandas/core/series.py in _sanitize_array(data, index, dtype, copy, raise_cast_failure)
       3274         if isinstance(data, np.ndarray):
    -> 3275             raise Exception('Data must be 1-dimensional')
       3276         else:

    Exception: Data must be 1-dimensional

    During handling of the above exception, another exception occurred:

    ValueError                                Traceback (most recent call last)
    <ipython-input-40-1aeb3139343b> in <module>()
    ----> 1 df = fit.to_dataframe()

    stanfit4anon_model_68f43e2dc885e62bb7e4c7117e91aa73_2098315045300022999.pyx in stanfit4anon_model_68f43e2dc885e62bb7e4c7117e91aa73_2098315045300022999.StanFit4Model.to_dataframe()

    ~/miniconda3/envs/stantmp2/lib/python3.6/site-packages/pystan/misc.py in to_dataframe(fit, pars, permuted, dtypes, inc_warmup, diagnostics)
       1200                 ss = ss.astype(dtypes[par])
       1201             if ss.shape[1] == 1:
    -> 1202                 df[par] = ss
       1203             else:
       1204                 par_flatnames = [

    ~/miniconda3/envs/stantmp2/lib/python3.6/site-packages/pandas/core/frame.py in __setitem__(self, key, value)
       2517         else:
       2518             # set column
    -> 2519             self._set_item(key, value)
       2520 
       2521     def _setitem_slice(self, key, value):

    ~/miniconda3/envs/stantmp2/lib/python3.6/site-packages/pandas/core/frame.py in _set_item(self, key, value)
       2582         """
       2583 
    -> 2584         self._ensure_valid_index(value)
       2585         value = self._sanitize_column(key, value)
       2586         NDFrame._set_item(self, key, value)

    ~/miniconda3/envs/stantmp2/lib/python3.6/site-packages/pandas/core/frame.py in _ensure_valid_index(self, value)
       2564                 value = Series(value)
       2565             except:
    -> 2566                 raise ValueError('Cannot set a frame with no defined index '
       2567                                  'and a value that cannot be converted to a '
       2568                                  'Series')

    ValueError: Cannot set a frame with no defined index and a value that cannot be converted to a Series

braaannigan:
Thanks @ahartikainen 

When you set the first column of the dataframe you also set the index. In this case pandas wants a vector that has shape (4000,) rather than shape (4000,1).  I've addressed this by selecting the first column of ```ss``` to make sure it has the right shape.  Both the test case and 8 schools should work now.

ahartikainen:
Looks great. I will test this today and if everything is fine, we can go and merge.

ahartikainen:
Hi, I tested the code and (almost) everything looks great.

I have one last request. Currently, when calling with `inc_warmup` it is quite hard to separate warm-up and non-warm-up samples. So if `inc_warmup` is True, add one column to indicate if sample is from warm-up or not (`bool`?)

After that, this PR is good to go.

ahartikainen:
Also, I was thinking that maybe we should use ints instead of bools in the dataframe.

Calling `df.describe()` will drop all the boolean columns. Even `df.describe(include='all')` does not work as expected (or did I have some `nans` in that column?, probably not).

All the normal boolean stuff will work normally with int column.

@braaannigan can you test and give your opinion about it?

braaannigan:
I've changed the warmup column to be an int and it now appears in the output of df.describe().  I've modified the tests and played around with a few configurations including with generated quantities and everything looks good.

braaannigan:
I'm having problems getting the tests to run on my local machine. When I try something like
```
py.test tests/
```
from the command line I get a string of error messages for each file in tests saying ``` ImportError: No module named _api```

Any suggestions as to how this should be done?

ahartikainen:
We use unittest?

braaannigan:
Lol, that's true!

braaannigan:
Do I need to update whats-new.rst as well?  What version number would it be?

ahartikainen:
No need to update it. We update it usually before release.

ahartikainen:
Looks good, tested and everything works! 

Makes working with data really nice, especially all the `df.groupby` magic that is enabled by this PR.

@braaannigan thank you for the PR and all your patience with our dev-process!

I will merge this now.

braaannigan:
Great, thanks for your patience @ahartikainen!

PR #426, opened 2018-02-05
Stop shuffling extracted samples

Praznat:
#### Summary:
When calling `extract`, the default argument `permuted=True` returns a parameter-keyed dictionary of samples in shuffled order. When `permuted=False`, samples are in the order they were sampled in, but the result is not longer a `dict` object, so the samples are harder to retrieve for a desired parameter. Traceplots also call `extract` with default arguments, so it is impossible to detect autocorrelation in the samples just by looking at the plots. The plan for PyStan3 already seems to be getting rid of `extract(permuted=True)`, as discussed in this thread: https://github.com/stan-dev/rstan/issues/345.

#### Intended Effect:
Extract parameter samples in the same order they were sampled, rather than in permuted order, while leaving intact the ability to retrieve them through a `dict` object keyed by parameter names. This will make `permuted=True` just arrange samples into a `dict` rather than actually shuffling them ("permuted" might not be the right word to use anymore).

#### How to Verify:
Fit a model, then call `extract()` and `extract(permuted=False)`. The first result should be a `dict` keyed by parameter names with values containing the parameter samples after the warmup period. The latter should be an array with parameters in the same order as `flatnames`, optionally containing the warmup samples. Both the former and the latter should have the same samples in the same order after the warmup period for any given parameter.

#### Side Effects:
None

#### Documentation:
The current documentation at http://pystan.readthedocs.io/en/latest/api.html says samples will be permuted by calling `permuted=True`. It is not clear from the documentation how they were being permuted, and it will still not be clear that they will now be in the same order they were sampled in. It may make more sense to call this argument `return_dict` rather than `permuted`.

#### Copyright and Licensing

Revionics

### DIFF
--- a/pystan/_misc.pyx
+++ b/pystan/_misc.pyx
@@ -9,7 +9,6 @@ cpdef get_kept_samples(int n, dict sim):
     """See documentation in misc.py"""
     cdef int i, j, num_chains, num_samples, num_warmup, ss_index, s_index
     cdef double[:] s, ss
-    cdef long[:] perm
     num_chains = sim['chains']
     nth_key = list(sim['samples'][0]['chains'].keys())[n]
 
@@ -19,11 +18,10 @@ cpdef get_kept_samples(int n, dict sim):
 
     ss = np.empty((num_samples - num_warmup) * num_chains)
     for i in range(num_chains):
-        perm = sim['permutation'][i]
         s = sim['samples'][i]['chains'][nth_key]
         for j in range(num_samples - num_warmup):
             ss_index = i * (num_samples - num_warmup) + j
-            s_index = num_warmup + perm[j]
+            s_index = num_warmup + j
             ss[ss_index] = s[s_index]
     return ss
 


ahartikainen:
I'm going to update the plotting code with recently merged PR. So autocorrelation issue will be handled in the future.

We could update our extract method and still keep the permuted kw. 
I'm not sure if removing permuted would break anything.

@ariddell what do you think?

ariddell:
I'll try to look closer at this tomorrow.

What seems important to me is (1) no breaking backward compatibility and
(2) consistency with RStan.

ahartikainen:
Hi, PyStan 3 is going to fix this so let's not merge this.

With the current codebase you can call `.extract` with `permuted=True` given pars is not `None`.

    draws = fit.extract(pars=fit.model_pars, permuted=False)

Or with `lp__`

    pars = list(fit.model_pars) + ["lp__"]
    draws = fit.extract(pars=pars, permuted=False)

We are going to remove plotting code from PyStan and start to use ArviZ library.

see 
https://github.com/arviz-devs/arviz
http://arviz-devs.github.io/arviz/

PR #438, opened 2018-02-20
#include external files

ahartikainen:
#### Summary:

Enable to read external files

#### Intended Effect:

`#include "stanfile.stan"` should read filecontents from external file.

#### How to Verify:

Create models where parameters are defined inside the external files

#### Side Effects:

Implements `include_paths`, `filename` and `allow_undefined` parameters for stanc. Check if we need to give those values defaults, in case of anyone is using the stanc.hpp file. Otherwise, defaults are given in Python stanc-function.
 
#### Documentation:

Needs updating. Comment how to do this correctly.

#### Reviewer Suggestions: 

@ariddell 
@bob-carpenter 

#### Copyright and Licensing

Ari Hartikainen


ahartikainen:
This still probably needs `os.path.abspath` to python side (pystan/api.py in the list comprehension).

pystan/stanc.hpp
```diff
@@ -38,12 +43,15 @@ int stanc(std::string model_stancode, std::string model_name, PyStancResult& res
 
   std::string mcode_ = model_stancode;
   std::string mname_ = model_name;
+  bool aundefined_ = allow_undefined;
+  std::string fname_ = filename; 
+  std::vector<std::string> ipaths_ = include_paths;
```
ariddell: Are any of these definitions necessary?

ariddell:
Thanks for this @ahartikainen .

For future reference, the way Stan C++ reaches into the filesystem is not, IMHO, helpful: https://github.com/stan-dev/stan/issues/2351

pystan/stanc.hpp
ahartikainen: Probably not, I just kept the original structure

ahartikainen:
True, it would be more robust if we read the files on a python side and then we could send a vector of keys (#include key) and a vector code for that key.

Now I'm not even sure that will work for Windows with '\' on a path, or even paths with spaces.

The reading is done here https://github.com/stan-dev/stan/blob/develop/src/stan/io/program_reader.hpp

    /**
       * Returns the characters following <code>#include</code> on
       * the line, trimming whitespace characters.  Assumes that
       * <code>#include</code>" is line initial.
       *
       * @param line line of text beginning with <code>#include</code>
       * @return text after <code>#include</code> with whitespace
       * trimmed
       */
      static std::string include_path(const std::string& line) {
        int start = std::string("#include").size();
        while (line[start] == ' ') ++start;
        int end = line.size() - 1;
        while (line[end] == ' ') --end;
        return line.substr(start, end - start);
      }

      void read(std::istream& in, const std::string& path,
                const std::vector<std::string>& search_path,
                int& concat_line_num,
                std::set<std::string>& visited_paths) {
        if (visited_paths.find(path) != visited_paths.end())
          return;  // avoids recursive visitation
        visited_paths.insert(path);
        history_.push_back(preproc_event(concat_line_num, 0, "start", path));
        for (int line_num = 1; ; ++line_num) {
          std::string line = read_line(in);
          if (line.empty()) {
            // ends initial out of loop start event
            history_.push_back(preproc_event(concat_line_num, line_num - 1,
                                             "end", path));
            break;
          } else if (starts_with("#include ", line)) {
            std::string incl_path = include_path(line);
            history_.push_back(preproc_event(concat_line_num, line_num - 1,
                                             "include", incl_path));
            bool found_path = false;
            for (size_t i = 0; i < search_path.size(); ++i) {
              std::string f = search_path[i] + incl_path;
              std::ifstream include_in(f.c_str());
              if (!include_in.good()) {
                include_in.close();
                continue;
              }
              try {
                read(include_in, incl_path, search_path, concat_line_num,
                     visited_paths);
              } catch (...) {
                include_in.close();
                throw;
              }
              include_in.close();
              history_.push_back(preproc_event(concat_line_num, line_num,
                                               "restart", path));
              found_path = true;
              break;
            }
            if (!found_path)
              throw std::runtime_error("could not find include file");
          } else {
            ++concat_line_num;
            program_ << line;
          }
        }
        visited_paths.erase(path);  // allow multiple, just not nested
      }


      /**
       * Read the rest of a program from the specified input stream in
       * the specified path, with the specified search path for
       * include files, and incrementing the specified concatenated
       * line number.  This method is called recursively for included
       * files.  If a file is included recursively, the second include
       * is ignored.
       *
       * @param[in] in stream from which to read
       * @param[in] path name of stream
       * @param[in] search_path sequence of path names to search for
       * include files
       * @param[in,out] concat_line_num position in concatenated file
       * to be updated
       * @throw std::runtime_error if an included file cannot be found
       */
      void read(std::istream& in, const std::string& path,
                const std::vector<std::string>& search_path,
                int& concat_line_num) {
        std::set<std::string> visited_paths;
        read(in, path, search_path, concat_line_num, visited_paths);
      }
    };

ahartikainen:
`std::string f = search_path[i] + incl_path;`

Does this mean that the path needs to end with `'/'`. That is not the default on python `os`

pystan/stanc.hpp
ariddell: Let's get rid of them and see if works?

bob-carpenter:
Windows will take paths with `/` from C++, so that's not an issue as far as I know.

Yes, I think the include paths need to end with `/`.  That's bad on my part and we should fix that.  Would you mind creating an issue?  I can check if there's already a `/` or `\` at the end and add a separating `/` if not.

bob-carpenter:
Belay that last request---I added the issue myself so as not to forget:

https://github.com/stan-dev/stan/issues/2472

ahartikainen:
Weird things happen. Retested the code and it works after putting "/" to the end.

`myfunc.stan`

    real theta;

And then calling it in

    model = """
    parameters {
        #include myfunc.stan
        real y;
    }
    model {
        theta ~ normal(0,1);
        y ~ normal(33,13);
    }
    """

    path = './' # or ['./'] both work

    sm = pystan.StanModel(model_code=model, include_paths=path)

I put the check for `/` in python side for now and remove it after we have released the 2.18 (assuming that is fixed in Stan lib). Is that ok?

bob-carpenter:
Certainly OK by me---sorry I didn't make a better one in the first place!

ahartikainen:
Sorry, I went for the ninja-edit mode.

So is it still that it needs to be at the start of the line? After that there can be whitespace? Don't C++ have a `string.strip` function or something similar (remove whitespace start and end)

bob-carpenter:
I don't know about stripping.

Yes, it's still assuming line-initial includes.  Which now that I think about it, probably means RStan's just stripping them.  All these things are pretty easy to change.

ahartikainen:
I added a new `model_code_` variable that contains "stripped" `model_code`. This stripped version goes to C++ and original goes to the name if obfuscation is on and returned in the dictionary.

Stripping is done with some regex magic. This works for multiple include files.

One thing still is to "rename" the filename to PyStan specific way "Unknown PyStan file name" etc.

ariddell:
@ahartikainen I'm -1 on the "editing" of Stan program code. People need to trust that the Stan program code they provide is the Stan program code that gets compiled. Can we put the whitespace handling in Stan C++ instead?

ahartikainen:
That is ok. Code was getting quite hacky.

I can remove it. I also need to refactor some of the commits.

ahartikainen:
Ok, I redid the commits and changed some parts.

I removed all the `/` and `\\` handling from the python side. Let's assume that the user knows what she/he is doing.

I kept the default `include_path=['./']`.

ahartikainen:
Currently there can't be any whitespace before `#include` and after `...file.stan`.

    "\n#include external.stan\n"

Also users need to add `'/'` to the end of the filepath. 
Windows users need to try if `'\\'` works or not.

ahartikainen:
Ok, I have tested all the tricks and travis doesn't like it. On my local machine everything works (from single line code with explicit `\n` to multilinecode with ugly hack to remove whitespace.

Any ideas?

ahartikainen:
Yep, I have no idea why this is currently failing. 

The path should be fine. (it uses the same syntax as other test that read data from data-folder).
Tests are passing on my laptop (macbook pro). 
Is there something weird with C++ and paths on travis? @bob-carpenter

riddella:
If you don't figure it out, I can test this on my linux machine sometime in the next few days. If it's happening on Travis, I suspect it will happen locally for me.

ariddell:
If you don't figure it out, I'll try this out on linux tomorrow.

ahartikainen:
Great! If I have some time, I will try docker+ubuntu and report the results here.

This could be some problem with path were there is e.g. dot ("sdsdsd/python3.6/sdsdsd") or the default encoding is off somehow.

ariddell:
Looking at this now. Sorry it took me so long to get to it.

ariddell:
Very strange. It's not a period in the path. It seems like you need to add the '/' earlier. This works for me:

```python
    def test_stanc_include(self):
        model_code = """parameters {
#include external1.stan
#include external2.stan
        }
        model {
        y ~ normal(0,1);
        z ~ normal(0,1);
        }"""
        testdata_path = os.path.join(os.path.relpath(os.path.dirname(__file__)), 'data/')
        result = stanc(model_code=model_code, include_paths=[testdata_path])
        desired = sorted({"status", "model_cppname", "cppcode", "model_name", "model_code", "include_paths"})
        self.assertEqual(sorted(result.keys()), desired)
        self.assertEqual(result['status'], 0)
```

I put the trailing ``/`` in the ``os.path.join``. Very strange.

ahartikainen:
That seems really weird. Could that be some kind of Cython error?

Are you certain it was not the `os.path.relpath`, which worked?

ariddell:
Sorry, you're right. The non-absolute "data/" is the only one that fails for me. I think it's fine if we require absolute paths.

ahartikainen:
Nope travis still fails. I have no idea why. Could this be gcc / C++ problem?

Tests work on mac and desktop Linux.

ariddell:
It's not working on clang either, right? So it's not gcc.

First idea: it's a travis-ci container/filesystem issue. Add a pre-test using ``os.path.exists`` to make sure that Python (not Stan C++) can find the included files in the directory.

Second idea: try doing the path construction in an os-independent manner. So use os.sep and maybe try using "normpath", as recommended here ``https://stackoverflow.com/questions/2736144/python-add-trailing-slash-to-directory-string-os-independently`` (We should be doing this anyway, actually)

Third idea: mark the feature as experimental in the documentation. We could add some sort of check to see if the test is being run on travis and skip this particular test.

@bob-carpenter Could this possibly be a C++ IO issue with older versions of clang or gcc? Our tests pass on Linux and macOS locally.

ariddell:
I know what it is. This is Python packaging stuff. You need to add those
include files to the package_data_pats.

package_data_pats = ['*.hpp', '*.pxd', '*.pyx', 'tests/data/*.csv',
                     'lookuptable/*.txt']

should become

package_data_pats = ['*.hpp', '*.pxd', '*.pyx', 'tests/data/*.*',
                     'lookuptable/*.txt']

Because the test files are not obviously Python files, the normal
setuptools utilities do not include them. One has to explicitly add them.

I'm really sorry you had to deal with this!

(In defense of the Python packaging community, nobody thinks that this
way of doing things makes sense. There's lots of work happening right
now to improve the packaging tools.)

ahartikainen:
Ok, it finally works. In what part of the docs should we add example of usage?

Do we want to give a warning if somebody is going to use this?

I could add regex "whitespace remover"- function to help with the feature (in `.experimental`).

ariddell:
Let's just add a discussion of the issues to the docstring for ``stanc``. I suspect that people will search for ``include`` if they want details.

ahartikainen:
Do we merge this or wait for https://github.com/stan-dev/stan/issues/2472 ?

If we wait (assuming it is in Stan 2.18), we don't need the whitespace removal function in `.experimental`.

If we don't wait for it, then I can merge both of these.

ariddell:
Let's assume we don't make a release until 2.18 is merged. I don't think there's a strong reason to rush this. People who want the bleeding edge can always compile things themselves.

ahartikainen:
Yes, I agree on this, let's wait.

ahartikainen:
I think we can merge this and the helper function, because the whitespace fix is going to come 2.18++.

PR #439, opened 2018-02-20
Rename diagnostic function to use snake case.

ariddell:
Function names use snake case (e.g., ``lower_case_with_underscores``).
It's not specified explicitly by PEP8 but it is almost a universal
practice.


ahartikainen:
Good catch, I try to be more accurate in future.

Issue #133, opened 2014-12-30
bug yielding Rhat values of "nan"


jbloom:
I am using PyStan to do MCMC over some variables in a simplex (Dirichlet prior). I check for convergence by looking at the Rhat value for each variable in the simplex. Occasionally, a few of the variables in the simplex have an Rhat of "nan." Even when I repeatedly re-run the MCMC with more steps, these variables keep having an Rhat of "nan." This is despite the fact that other variables in the simplex will have Rhat values of very close to one (<1.02) and the effective sample size (n_eff) will be >1000 even for the variables returning "nan" for Rhat.

I am speculating that there is some bug in the Rhat calculation that ends up returning "nan" for pathological values of certain variables?

jbloom:
Hi Bob,

I don't think that the problem is just that the variable is unchanged over the MCMC. For instance, at the bottom of this e-mail I have printed out the Rhat, n_eff, 25%, and 75% interval values for the elements of the simplex (Dirichlet prior). You can see that about half the variables in the simplex have a Rhat of nan. However, the n_eff is still quite large and the 25% and 75% interval values are different, so the simplex element is sampling distinct values during the MCMC. 

In general, simplex elements with small numerical values tend to be more likely to produce nan, although there is not a complete correlation.

Note that this is a fairly rare issue. I am running large numbers of MCMC chains with simplexes and monitoring convergence by Rhat, and only rarely does this problem arise. However, it definitely causes a serious problem when trying to write programs that are automatically using Rhat to check for convergence.

Here are the Rhat, n_eff, 25%, and 75% values for the elements of the simplex pir taken from the fitsummary produced by pystan:

For variable pir[0], values are:
	Rhat = 1.00209
	n_eff = 1123
	25% = 0.29137
	75% = 0.329987
For variable pir[1], values are:
	Rhat = nan
	n_eff = 4701
	25% = 4.50462e-05
	75% = 0.00021267
For variable pir[2], values are:
	Rhat = nan
	n_eff = 3336
	25% = 0.000134368
	75% = 0.000385033
For variable pir[3], values are:
	Rhat = 1.00483
	n_eff = 1991
	25% = 0.00107831
	75% = 0.00292147
For variable pir[4], values are:
	Rhat = nan
	n_eff = 2997
	25% = 0.00027064
	75% = 0.000641184
For variable pir[5], values are:
	Rhat = nan
	n_eff = 2185
	25% = 0.00190983
	75% = 0.00260189
[...]

PR #442, opened 2018-02-27
Fix spurious NaNs in split_potential_scale_reduction

jjramsey:
The function `split_potential_scale_reduction()` in `_chains.pyx`, which calculates Rhat, has a problem where it compares a value to another arbitrary value that is presumed to be "close enough" to zero. Here are the lines in question:
```
if var_within < pystan.constants.EPSILON:
    srhat = float('nan')
else:
    srhat = sqrt((var_between/var_within + n_samples / 2 - 1)/ (n_samples / 2))
```
The problem is that it is quite possible to get a value of `var_within` less than `pystan.constants.EPSILON` (which is only 1e-6, rather than machine epsilon) even if the parameter in question isn't pathological. (I've had this happen, so it isn't just a hypothetical. Also see some of the discussion in the old issue #133.)

Furthermore, since `cdivision` is `True` for `split_potential_scale_reduction`, even if `var_within` is zero, there shouldn't be a `ZeroDivisionError` exception, so there's no need for that comparison. In the case of division by zero, `split_potential_scale_reduction()` would produce a *non*-spurious Inf or NaN.

#### Copyright and Licensing

Please list the copyright holder for the work you are submitting (this will be you or your assignee, such as a university or company): James J. Ramsey

### DIFF
--- a/pystan/_chains.pyx
+++ b/pystan/_chains.pyx
@@ -229,10 +229,7 @@ def split_potential_scale_reduction(dict sim, int n):
     cdef double var_between = n_samples / 2 * stan_variance(split_chain_mean)
     cdef double var_within = stan_mean(split_chain_var)
 
-    if var_within < pystan.constants.EPSILON:
-        srhat = float('nan')
-    else:
-        srhat = sqrt((var_between/var_within + n_samples / 2 - 1)/ (n_samples / 2))
+    srhat = sqrt((var_between/var_within + n_samples / 2 - 1)/ (n_samples / 2))
     return srhat
 
 
--- a/pystan/_chains.pyx
+++ b/pystan/_chains.pyx
@@ -229,7 +229,7 @@ def split_potential_scale_reduction(dict sim, int n):
     cdef double var_between = n_samples / 2 * stan_variance(split_chain_mean)
     cdef double var_within = stan_mean(split_chain_var)
 
-    srhat = sqrt((var_between/var_within + n_samples / 2 - 1)/ (n_samples / 2))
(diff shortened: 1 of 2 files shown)


ariddell:
Thanks for the PR.

This is related to #133. I recall introducing the EPSILON check because something was raising a Python or Cython exception.

There's some git magic to figure out when the code got introduced.

jjramsey:
Did you find out when that EPSILON check was introduced, and if it was before or after cdivision was set to true?

pystan/_chains.pyx
```diff
@@ -229,10 +229,7 @@ def split_potential_scale_reduction(dict sim, int n):
     cdef double var_between = n_samples / 2 * stan_variance(split_chain_mean)
     cdef double var_within = stan_mean(split_chain_var)
 
-    if var_within < pystan.constants.EPSILON:
-        srhat = float('nan')
-    else:
-        srhat = sqrt((var_between/var_within + n_samples / 2 - 1)/ (n_samples / 2))
+    srhat = sqrt((var_between/var_within + n_samples / 2 - 1)/ (n_samples / 2))
```
ariddell: Needs consistent spacing around division operators.

ariddell:
Looks like it was before cdivision was set to true. So this change makes sense.

ariddell:
Thanks. We should probably remove the constants.EPSILON entirely since we're not using it anywhere else.

ahartikainen:
Looks good.

ariddell:
Thanks @jjramsey . Glad to have this fixed.

PR #457, opened 2018-04-25
Use pars kwarg in vb

ahartikainen:
#### Summary:

Fixed pars argument bug. This enables the use of pars argument.

#### Intended Effect:

Possibility to use pars argument with vb.

#### How to Verify:

`model.vb(pars=['theta']`

#### Side Effects:

None

#### Documentation:

No edits needed.

#### Copyright and Licensing

Ari Hartikainen


ahartikainen:
Now the parameter names are the same as in sampling. (`a.1.1` --> `a[1,1]`)

ahartikainen:
Currently sample params in vb results contains `lp__` which is array 0 (sample params are iid samples from the marginal distribution, so that makes sense (?)). Should we filter it, so `sample_params` would have the same amount of variables as does mean variable.

It is filtered manually from the results.

Also we should remove the first item from the sample params which is the mean value.

And last, should we transform variables to numpy array?

cc @ariddell

ariddell:
My sense is that we should do whatever is simplest or whatever R does. Users of ``vb`` are going to be advanced users; they will read the documentation carefully.

ahartikainen:
I added `mean_par_names` key, which is `sampler_param_names` with `lp__` removed. Now it matches the `mean_pars` as `sample_param_names` matches `sample_params`.

`sample_params` now contains only the posterior draws, so the first value is removed. This is currently implemented on the Python side. If needed, I can do it also in the C++ side if needed.

ahartikainen:
I'm happy with this. 

pars argument is working.

Now mean and posterior sample are in different objects and explicitly named.

Names follow common syntax (same as sampling).

I will merge this on monday if there is still something to be done.

Issue #465, opened 2018-06-09
Logging behavior is altered when pystan is loaded


collincburton:
#### Summary:
If I load a logger with the basic config and say that it should write to a file, after I import pystan, it will write all of my error messages to the console.

#### Description:
Logging doesn't appear to be implemented correctly in pystan as it is altering the behavior of the logging functionality I have built into my script.

#### Reproducible Steps:
If you run this code with the import pystan either included or excluded, it will change the behavior of the logging.

```
import logging
import pystan


logging.basicConfig(filename='log_test.log', level=logging.DEBUG, format='%(asctime)s - %(levelname)s - %(message)s')

logging.debug("test debug message")
logging.info("Info message")
logging.warning("Warning message")
```

#### Current Output:
logging output is written to the console

#### Expected Output:
Logging output should be written to a file.

#### PyStan Version:
2.17.1.0

#### Python Version:
3.6.4

#### Operating System:
MacOS 10.13.5

collincburton:
It looks like it's a problem with __init__.py at line 18
```
if len(logger.handlers) == 1:
    logging.basicConfig(level=logging.INFO)
```
Per the logging docs (https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)

> Note It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

When you add the basicConfig, you are adding a lot of different handlers and it is interfering with the desired logging behavior of the end user.

riddell-stan:
This is how things work in PyStan. We use the logging facility to print messages. PyStan is not really a library, it's an application. I realize this is an idiosyncratic choice and we might want to consider printing our messages for the users to stderr with the `print` function.

bob-carpenter:
Sorry this is getting off topic, but I think it would be helpful to think of PyStan as a library in the same sense as RStan.  Other Python packages might want to import it and use it.  It's not a low-level library like NumPy, but some of the same considerations about controlling I/O apply.  It's been an ongoing struggle in RStan to get truly silent output, for example.

riddell-stan:
I think this is on topic! We should add documentation about how to 
silence the pystan (logging) output entirely. It shouldn't be difficult. 
[...]

PR #466, opened 2018-06-12
Remove basicConfig logger

collincburton:
Adding the basicConfig call initiates additional handlers, which can interfere with logging of downstream users.

#### Summary:
Currently, when pystan is imported, it can interfere with logging of end users because a basicConfig method is called in __init__.py
#### How to Verify:
See #465 for verification code. Currently, the test case will not produce a log file. After the pull request is merged, a log file will be created with appropriate logging info.

#### Documentation:
Per logging documentation that basicConfig should not be called in a library:

> Note It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

[https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)
#### Copyright and Licensing

### DIFF
--- a/pystan/__init__.py
+++ b/pystan/__init__.py
@@ -14,8 +14,6 @@ from pystan.lookup import lookup
 
 logger = logging.getLogger('pystan')
 logger.addHandler(logging.NullHandler())
-if len(logger.handlers) == 1:
-    logging.basicConfig(level=logging.INFO)
 
 # following PEP 386
 __version__ = '2.17.1.0'
--- a/pystan/__init__.py
+++ b/pystan/__init__.py
@@ -14,6 +14,8 @@ from pystan.lookup import lookup
 
 logger = logging.getLogger('pystan')
 logger.addHandler(logging.NullHandler())
+if len(logger.handlers) == 1:
+    logger.setLevel(logging.INFO)
 
 # following PEP 386
 __version__ = '2.17.1.0'


ahartikainen:
Instead of removing those lines, should it just change the `logger` instance?

    if len(logger.handlers) == 1:
        logger.setLevel(logging.INFO)

riddell-stan:
I think @ahartikainen is on the right track, if you add a handler in advance, basicConfig will not be called.

A related issue that I'll raise here. How should we be printing information for the user? We will have users working in at least the following three environments:

- console
- IPython / REPL
- Jupyter Notebook

collincburton:
I made the change that @ahartikainen suggested and it works with my local test cases. I can still get debug messages with my code even though the module is set to INFO.

Re: @riddell-stan I'm adding handlers in advance with the test code in #465 via a call to basicConfig in the end user script and the output is still be changed via the pystan library.

riddell-stan:
I believe this will prevent any information from being printed to the typical user. Calling `basicConfig` adds a default StreamHandler so users will be able to see progress reports and various other bits of information. setLevel does not add a StreamHandler.

If we want to stop using logging and switch to using `print` statements, we could talk about that. But perhaps the discussion board would be a better place to do so?

See also my note on https://github.com/stan-dev/pystan/issues/465.

riddell-stan:
Looks like pymc 3 does exactly what we do:

https://github.com/pymc-devs/pymc3/blob/850a2a7eb62c505cc7e54371b034d0afdb56262a/pymc3/__init__.py

ahartikainen:
Ok, I think we could do the same.

1. Get the "pystan" logger
2. Check if logger has any handler
3. If it doesn't have any create a new streamhandler, set it to `INFO` and add to handlers.

This way if users want to handle loggers they have this option. E.g. saving to file on server.

ahartikainen:
So something like this

    logger = logging.getLogger('pystan')
    if not len(logger.handlers):
        null_handler = logging.NullHandler()
        logger.addHandler(null_handler)
        logger.setLevel(logging.INFO)

ahartikainen:
@riddell-stan do you have opinion on this?

Based on this http://docs.python-guide.org/en/latest/writing/logging/

we could even just add nullhandler

"It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers."

    import logging
  
  logging.getLogger(__name__).addHandler(logging.NullHandler())

riddell-stan:
I'm not persuaded that there is a bug that needs fixing here. I *think* what we are doing right now is fine because the typical use of pystan is as an application, not a library. (The python docs refer to library modules.) We're doing the same thing as pymc, as far as I can tell.

Maybe we should move this to the discussion forum.

As for this:

```
logger = logging.getLogger('pystan')
if not logger.handlers:
    null_handler = logging.NullHandler()
    logger.addHandler(null_handler)
    logger.setLevel(logging.INFO)
```

Users of the app would get no messages, right? No info, no warning messages.

Also, what you describe in https://github.com/stan-dev/pystan/pull/466#issuecomment-405018144 , I think that's exactly what we do. The call to ``basicConfig`` does add a StreamHandler. (I know the code in ``__init__.py`` and in pymc look different, but I think they do exactly the same thing.) Here are the docs for basicConfig file:///usr/share/doc/python3.6-doc/html/library/logging.html?highlight=basicconfig#logging.basicConfig

ahartikainen:
True, I was thinking of StreamHandler.

But basically, if the user wants to log to file, the user should config pystan logger so in that case the basicConfig is not triggered.

We can close this PR and create another one for the correct instructions how to do it. Am I right?

ahartikainen:
Closing this PR. The correct instructions for logger-handling are in #478.

PR #478, opened 2018-07-18
[Docs] Logging

ahartikainen:
#### Summary:

Add instructions for logging module. Contains a simple example how to save output to a file.

#### Intended Effect:

Enable user to output messages to file.(or any other destination)

#### How to Verify:

Save output to logfile.

#### Side Effects:

Disables output if users wants that.

#### Documentation:

Added logging section.

#### Copyright and Licensing

Ari Hartikainen

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -26,6 +26,7 @@ Documentation
    differences_pystan_rstan
    api
    conversion
+   logging
    windows
 
 Stan documentation
--- a/doc/logging.rst
+++ b/doc/logging.rst
@@ -28,6 +28,11 @@ To redirect all messages only to a file.
 
     import logging
     logger = logging.getLogger("pystan")
+    
+    # add root logger (logger Level always Warning)
+    # not needed if PyStan already imported
+    logger.addHandler(logging.NullHandler()) 
+    
     logger_path = "pystan.log"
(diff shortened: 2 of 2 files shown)


ahartikainen:
Other format are also ok.

I still need to test the filehandler is working correctly.

ahartikainen:
I believe there should be NullStream for the root logger, so then the setlevel would work. Apparently root is always set to Warning.

riddell-stan:
You mean ``NullHandler``? We add that in ``__init__.py``.

https://github.com/stan-dev/pystan/blob/6c557cf53d833939651a2bb0cfe89b4f166db709/pystan/__init__.py#L16

ahartikainen:
Yes, `NullHandler` for the example (before pystan is imported)

riddell-stan:
This all looks good. I wish we had a mature package which uses pystan as a library -- then we'd know what to do based on experience.

(My suspicion is that an application using a pre-compiled model and using pystan purely as a library would want to silence all PyStan logging output.)

ahartikainen:
Now, this is working as intended. 

If `verbose=True` for `StanModel` it will still print stuff from Cython to stdout. I don't know if there is a way to grab that. 
Some wrapper for function that would grab prjmts, warnings and exceptions and use logging to output them (and then raise exception if needed)

Issue #463, opened 2018-05-24
cvodes library replaced by sundials


ariddell:
See https://github.com/stan-dev/math/pull/832

Most of the needed changes may be read off of the updated cmdstan Makefile.

riddell-stan:
Fixed by #480

PR #480, opened 2018-07-20
Update Stan to tag v2.18.0, update include paths

### LINKED ISSUE #479: PyStan should use Stan 2.18 release
Stan 2.18 is released. PyStan should update.

Other 2.18 related issues are in the 2.18 Milestone https://github.com/stan-dev/pystan/milestone/12

riddell-stan:
Closes #479, #463

### DIFF
--- a/pystan/model.py
+++ b/pystan/model.py
@@ -259,8 +259,8 @@ class StanModel:
             os.path.join(pystan_dir, "stan", "src"),
             os.path.join(pystan_dir, "stan", "lib", "stan_math"),
             os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "eigen_3.3.3"),
-            os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "boost_1.64.0"),
-            os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "cvodes_2.9.0", "include"),
+            os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "boost_1.66.0"),
+            os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "sundials_3.1.0", "include"),
             np.get_include(),
         ]
 
@@ -292,9 +292,14 @@ class StanModel:
                 '-ftemplate-depth-256',
                 '-Wno-unused-function',
                 '-Wno-uninitialized',
+                '-std=c++11',
             ]
             if platform.platform().startswith('Win') and build_extension.compiler in (None, 'msvc'):
-                extra_compile_args = ['/EHsc', '-DBOOST_DATE_TIME_NO_LIB']
+                extra_compile_args = [
+                    '/EHsc',
+                    '-DBOOST_DATE_TIME_NO_LIB',
+                    '/std:c++11',
+                ]
 
         distutils.log.set_verbosity(verbose)
         extension = Extension(name=self.module_name,
--- a/pystan/tests/test_extra_compile_args.py
+++ b/pystan/tests/test_extra_compile_args.py
@@ -11,6 +11,7 @@ class TestExtraCompileArgs(unittest.TestCase):
             '-ftemplate-depth-1024',
             '-Wno-unused-function',
             '-Wno-uninitialized',
+            '-std=c++11',
         ]
         model_code = 'parameters {real y;} model {y ~ normal(0,1);}'
         model = pystan.StanModel(model_code=model_code, model_name="normal1",
--- a/setup.py
+++ b/setup.py
@@ -101,8 +101,8 @@ from distutils.extension import Extension
 stan_include_dirs = ['pystan/stan/src',
                      'pystan/stan/lib/stan_math/',
                      'pystan/stan/lib/stan_math/lib/eigen_3.3.3',
-                     'pystan/stan/lib/stan_math/lib/boost_1.64.0',
-                     'pystan/stan/lib/stan_math/lib/cvodes_2.9.0/include']
+                     'pystan/stan/lib/stan_math/lib/boost_1.66.0',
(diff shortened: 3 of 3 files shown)


ahartikainen:
2.18 now can contain thread_local in cpp code? 

And if no `STAN_THREADS` is found in environmental variables it sets it automatically to 1 (does this happen on python side or C++ side)?

If user manually sets `STAN_THREADS` environmental variable, will it use that or is there a possibility that we could check if user has set that env-variable (and output warning that it is not tested)?

ahartikainen:
Parallel sampling works on Windows (automatic on Jupyter notebook, `if __name__ == "__main__"` block on python with a file). Or at least with mingw. 

I was thinking more of a computer with more than ~4 cores (let's say 20 or more) where one could use `thread_local` to parallelize large matrix computations (e.g. GP). Or do we need to wait MPI before that can happen?

ahartikainen:
Ok, I read some forum discussions. This could speed up stuff with usage of `map_rect` function?  So it could be beneficial for large number of cpu cores.

Matrix calculations are probably going for MPI/GPU.

bob-carpenter:
That's right---map_rect parallelizes its computation over as many cores as you give it.  The map step is applying the function in parallel, then there's a reduce step to put it back into the single autodiff expression graph being managed for a single log density eval.  There's nothing built into the matrix ops themselves (yet, anyway), but it becomes trivial to parallelize things like the X * beta data times parameter vector multiply you get in GLMs. 

When GPUs land soon, that is attacking the parallelization of the inner matrix algorithms, initially multiplication and Cholesky factorization.

bob-carpenter:
I believe the performance hit per chain for multi-threading is between 10 and 20% because the autodiff memory arena needs to be declared thread local, so each thread gets it own.  That adds overhead to every access.

ahartikainen:
`tests/test_extra_compile_args` needs c++11 argument.

ahartikainen:
Also windows needs one of these:

    /std:c++11
    /std:c++14
    /std:c++latest

https://msdn.microsoft.com/en-us/library/mt490614.aspx

riddell-stan:
But mingw is going to crash if we turn on thread_local, right (because 
it doesn't support C++11 threads, IIRC)? Isn't that going to bother 
Windows users a great deal?

Is thread_local required for map_rect? In any event, I think the 
guaranteed crash dialog on Windows is a significant concern, so I don't 
want to turn STAN_THREADS on by default until that's resolved.

ahartikainen:
Yes, let's not put it as a default. 

User must define env variable `STANTHREADS` and if it's defined let's output warning. Also if on Windows we can warn about the crash.

riddell-stan:
Let's deal with this in a separate PR. For example, we'll need at least 
one test which requires STAN_THREADS to be defined (that is, in C++).

riddell-stan:
@ahartikainen any ideas about the Windows errors?

```
stanfit4normal1_106238978.obj : error LNK2001: unresolved external symbol "public: double & __thiscall stan::math::internal::empty_broadcast_array<double,double>::operator[](int)" (??A?$empty_broadcast_array@NN@internal@math@stan@@QAEAANH@Z)
stanfit4normal1_106238978.obj : error LNK2001: unresolved external symbol "public: double & __thiscall stan::math::internal::empty_broadcast_array<double,int>::operator[](int)" (??A?$empty_broadcast_array@NH@internal@math@stan@@QAEAANH@Z)
C:\Users\appveyor\AppData\Local\Temp\1\tmpfjxy4n_m\stanfit4normal1_106238978.cp35-win32.pyd : fatal error LNK1120: 2 unresolved externals
```

ahartikainen:
No idea. Linker is somehow failing. Could you try C++latest?

It might be related to Sundials?

riddell-stan:
You mean use a more recent version of MSVC? Is this easy to do on appveyor?

ahartikainen:
No

This: `/std:c++latest

Can we somshow increase verbosity?

ahartikainen:
Apparently boost is too old for the latest. I may need to do some hand testing.

Can they compile this succesfully on Stan/math

bob-carpenter:
> Apparently boost is too old for the latest.

Which Boost is too old for what?  You mean we're not up to a recent enough version of Boost?

> Can they compile this succesfully on Stan/math

Compile what on stan/math?

bob-carpenter:
Yes, the continous integration tests Windows, but I think it's only testing some versions of gcc.  

We're actually going to reduce the number of test configurations we have---there's been an ongoing discussion on Discourse and it was just resolved.  

I have no idea where that error's coming from.  It's just a random template metaprogram that's part of the vectorization in the math lib.

We definitely need C++11 and as much of C++14 is supported by the g++ version in Rtools (that's our upper limit in order to be able to run under R).

riddell-stan:
Related, should we perhaps update the PyStan windows docs to state clearly that PyStan on Windows with MSVC is currently *unsupported* (since Eigen functions don't work, and, with 2.18 it's not clear if anything apart from stanc works) and PyStan on Windows with mingw-64 is supported.

Current docs: https://pystan.readthedocs.io/en/latest/windows.html

ahartikainen:
I tried to do recursive clone from your fork, but git checkout keeps the old libs. 

Boost is still 1.64 and no sundials. I have to investigate how to do cloning correctly.

ahartikainen:
I was able to clone your repo. It might be a path problem. Either some sundials path is missing or in "wrong" order so msvc linker don't find it.

It compiles pystan correctly, so I will investigate that.

ahartikainen:
Yes, I think msvc is not going to work.

Merge this and I will update stuff for mingw (appveyor + model.py + setup.py)

riddell-stan:
Sounds good to me. Thanks for offering to fix this.

PR #491, opened 2018-08-05
Bugfix/sampling

ahartikainen:
#### Summary:

WIP: Fix array of zeros with "Fixed_param"

Fix Errors raised with iter=1

Use hmc diagnostics only for nuts and hmc (automatic check)

Show truncated summary with __repr__. Helps if user has 100k variables.
    WIP: Number parameters?

#### Side Effects:

__repr__ may differ from __str__

#### Documentation:

WIP: usage for custom __repr__

#### Reviewer Suggestions: 

@riddell-stan 

#### Copyright and Licensing

Ari Hartikainen


ahartikainen:
Okey, I have implemented a `\_\_repr\_\_` so it will cut to 100 parameters and outputs a warning (to logger and also on the printed table).

I force warmup to 0 if algorithm="Fixed_param". There was no easy way to keep warmup as zero in the correct location. I think this will work as users think it should.

    warmup --> 0
    iter == num_samples

Also diagnostics had a problem if there was only 1 variable --> len needed array. So I changed the way to calculate sample (ndraws) size.

riddell-stan:
Seems like a good solution!

If it works, let's merge it. Remember to squash the little wip / fixup commits, if you don't mind. In theory, this will be part of the coding style we're going to borrow from Astropy (see http://docs.astropy.org/en/stable/development/workflow/development_workflow.html#rebase-but-only-if-asked )

ahartikainen:
Yes, I always do squash merge :

riddell-stan:
Sorry, I should have said rebase rather than squash. The idea, if I understand things correctly, is just eliminate the trial-and-error commits. The goal isn't to create 1 commit, it's typically just to reduce things from 30 commits to 1-3 commits. One would do this with ``git rebase -i`` typically. This talks about the procedure but doesn't give much detail: http://docs.astropy.org/en/stable/development/workflow/development_workflow.html#how-to-squash

ahartikainen:
That was quite easy. Let's see if these get run and merge after it.

riddell-stan:
Perfect! Much easier to read.

PR #497, opened 2018-08-12
DOC: update release notes for 2.18

riddell-stan:
If there are no objections, PyStan 2.18.0.0 will be released tomorrow, 2018-08-13.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -6,13 +6,13 @@
  What's New
 ============
 
-v2.18.0.0 (TBD)
+v2.18.0.0 (16. Aug 2018)
 ========================
 - Update Stan source to v2.18.0 (`release notes <https://github.com/stan-dev/stan/releases/tag/v2.18.0>`_)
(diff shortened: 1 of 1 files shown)


ahartikainen:
Hi, I'm making small addition to lib still.

DOC: How to enable threading + misc.py function for default gxx.

I will add that this is not working with Windows (tested and it will crash). Apparently, this does work with CmdStan where it probably fails silently at exit.

ahartikainen:
Hi, should we add info about the `pystan.diagnostics`?

    - `pystan.diagnostics` was implemented. Thanks @jjramsey

riddell-stan:
I'll add the diagnostics note right now.

I'll wait for the threads PR then. We could also release now and make another release in a few weeks when your patch is done.

ahartikainen:
I think we can merge the patch now (cancel the ci) and then wait for this to compile ready?

It is just a small function and docs. We can send people there when they start to ask how to use threads.

riddell-stan:
There's no harm in waiting. We can make a release when it's ready.

PR #498, opened 2018-08-12
Instructions for usage of STAN_THREADS

ahartikainen:
#### Summary:

Enable in-chain threading for advanced users.

#### Intended Effect:

Run in-chain parallel with threads.

#### How to Verify:

Run example

#### Side Effects:

Not supported on Windows.

#### Documentation:

Documentation updated.

#### Reviewer Suggestions: 

@riddell-stan 

#### Copyright and Licensing

Ari Hartikainen


ahartikainen:
I somehow named the function as gxx_args

It uses now cxxflags. Sorry about the mistake.

riddell-stan:
I'm trying to figure out what we should say about this in the docs.

Ignoring Windows for the moment, there's no benefit (indeed, there's a small cost) to using threads for parallelism. The only people who need STAN_THREADS are people who want to use new functions which need it (e.g., ``map_rect``).

For Windows -- I thought we were forgetting about this until clang-cl came out. Doesn't gcc always crash?

ahartikainen:
I did say that it won't work on Windows. With httpstan it kind of works, but with PyStan it crashes quickly. No chance to save the samples.

I think we could mention that this is only for advanced users who know what they are trying to do.

riddell-stan:
I would think the natural way to define STAN_THREADS would be to pass ``['-pthread', '-DSTAN_THREADS']`` as the argument for ``extra_compile_args`` when compiling a model.

If this doesn't work it might be because there's a problem in the code right now. Whatever the user passes as ``extra_compile_args`` should get added (or prepended) to the default args.

riddell-stan:
What might be good here would be to actually exercise the extra_compile_args in the test ``text_extra_compile_args.py``. Right now I don't think we are actually testing it with a meaningful argument.

ahartikainen:
Currently extra_compile_args overwrites the defaults.

I made the default_cxx function to easily get the defaults which are more or less arbitrary / hard to remember.

Not sure if `pthread` are needed, they should be default for all compiler we use.(Am I missing somdthing here?)

riddell-stan:
Yeah, I think there's a problem with extra_compile_args. The name (i.e., 
the "extra") implies that it should *not* override the defaults but 
rather add to them. This is a bug! I'm happy to fix it.

In any event, we should definitely have a test which exercises 
``-DSTAN_THREADS``.

ahartikainen:
Should we add compile_args if user wants ro overwrite the default settings?

I think we can release 2.18 and fix this in 2.18.1 (or 2.19 which is apparently planned to be released in a few weeks before StanCon Helsinki)

riddell-stan:
@ahartikainen see #499 . As for allowing users to completely override the default known-good compiler options, I think we might wait until users request the feature.

And if we do decide to allow overriding the default options, we might want to see how RStan does it (especially what parameter name it uses).

(In fact, I'm tempted to propose removing ``extra_compile_args``. I don't think RStan allows it.)

ahartikainen:
Ok, I updated the docs based on the #499 and removed unnecessary default cxxflags from pystan/misc.py.

I think this change looks a bit better now.

I will add one test so that should fail until #499 gets merged and I can rebase this with develop.

riddell-stan:
@ahartikainen Just merged #499 .

PR #499, opened 2018-08-14
Append `extra_compile_args` to default arguments.

riddell-stan:
This was always the intended behavior. The parameter name
`extra_compile_args` implies that the arguments are in addition to the
default arguments. With luck, no users are relying on the old
(unintended) behavior.

See also #498

### DIFF
--- a/pystan/model.py
+++ b/pystan/model.py
@@ -291,29 +291,39 @@ class StanModel:
         # compile stan models with optimization (-O2)
         # (stanc is compiled without optimization (-O0) currently, see #33)
         if extra_compile_args is None:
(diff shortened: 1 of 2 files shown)


ahartikainen:
How will this affect for if user adds `-O3` argument and we have `-O2` ? Was it so that first seen arguments are used?

Code looks good to me.

riddell-stan:
Looks like ``gcc -O2 -O3 file.c`` works fine. I'm guessing it's the last 
`-O3` which wins. I modified the code so that the extra arguments come last.

ahartikainen:
LGTM

PR #502, opened 2018-08-16
Release 2.18.0.0

riddell-stan:
Bumps the version number. There's also a minor change to the README (which involves the version number).


ahartikainen:
Do we need to remove some tests so the last step has time to finish.

Upload step is going to overtime?

ahartikainen:
Yes. Currently the test arr run explicitly, so we can reduce them to e.g. basic only.

PR #515, opened 2018-08-31
use logger.warning

ahartikainen:
#### Summary:

Use logger.warning for warnings

#### Intended Effect:

Enable easier handling for warnings


ahartikainen:
Maybe we can restart appveyor or just merge?

PR #516, opened 2018-09-01
Reduce Windows testing

ahartikainen:
#### Summary:

Reduce testing by excluding test taking longer time.

#### Intended Effect:

Speed up the CI Appveyor running times.

#### How to Verify:

Running times before:

40-60+ (min)

After

30-40 (min)


ahartikainen:
If we want to lower, let's make another pr.

PR #518, opened 2018-09-13
Correct path to sundials source.

### LINKED ISSUE #517: cvodes broken
cvodes should use the current folder structure (Sundials).

I tried to fix this by following cmakelist in cvodes folder, but I couldn't get `STAN_SUNDIALS_FPRINTF` to link correctly.

https://github.com/stan-dev/math/blob/develop/lib/sundials_3.1.0/src/cvodes/CMakeLists.txt

`STAN_SUNDIALS_FPRINTF` is defined in here: https://github.com/stan-dev/math/blob/develop/lib/sundials_3.1.0/include/stan_sundials_printf_override.hpp

Based on the quick google apparently either distutils.Extension is "broken" and it might be fixed with setuptools. Other one was that the order of included libs should come after -o python_module.so

`extra_link_args` going to Extension does put stuff after the `.so` but I still couldn't get it to link.

riddell-stan:
Responds to library merge, see stan-dev/math#832 for details.

Closes #517


ahartikainen:
I guess we have a much harder problem at hand. Clang does not like `-std=c++11` for c files.

    error:invalid argument '-std=c++11' not allowed with 'C'

Also, in my tests with gcc I needed to add all the files Makefile before I got the step where the linked header was not found.

ahartikainen:
Maybe we should fix this before VI?

We need to compile c & c++ which is problem for clang++, can we do this two parts?

riddell-stan:
Good catch. We definitely do need to fix this. It *was* working fine at one point (2.16?). I never figured out what went wrong.

wds15:
The approach in cmdstan is to use the C++ compiler, but pass the argument `-x c` to it which makes the C++ compiler work just like a C compiler.

- What is the difficulty with the C compiler?
- Can you use the makefiles from cmdstan?
- Is there an easy way to reproduce this problem (if it is still there)?

ahartikainen:
I think we could hack the `build_ext`

```
def edit_build_ext(build_extension):
    import types

    def build_extension_cpp(self, ext):
        sources = ext.sources
        if sources is None or not isinstance(sources, (list, tuple)):
            raise DistutilsSetupError(
                  "in 'ext_modules' option (extension '%s'), "
                  "'sources' must be present and must be "
                  "a list of source filenames" % ext.name)
        sources = sorted(sources)

        ext_path = self.get_ext_fullpath(ext.name)
        depends = sources + ext.depends
        if not (self.force or newer_group(depends, ext_path, 'newer')):
            log.debug("skipping '%s' extension (up-to-date)", ext.name)
            return
        else:
            log.info("building '%s' extension", ext.name)

        sources = self.swig_sources(sources, ext)

        extra_args = ext.extra_compile_args or []

        macros = ext.define_macros[:]
        for undef in ext.undef_macros:
            macros.append((undef,))

        objects = self.compiler.compile(sources,
                                         output_dir=self.build_temp,
                                         macros=macros,
                                         include_dirs=ext.include_dirs,
                                         debug=self.debug,
                                         extra_postargs=extra_args,
                                         depends=ext.depends)

        self._built_objects = objects[:]

        if ext.extra_objects:
            objects.extend(ext.extra_objects)
        extra_args = ext.extra_link_args or []

        # Detect target language, if not provided
        language = ext.language or self.compiler.detect_language(sources)
        if language == "c":
            language = "c++"

        self.compiler.link_shared_object(
            objects, ext_path,
            libraries=self.get_libraries(ext),
            library_dirs=ext.library_dirs,
            runtime_library_dirs=ext.runtime_library_dirs,
            extra_postargs=extra_args,
            export_symbols=self.get_export_symbols(ext),
            debug=self.debug,
            build_temp=self.build_temp,
            target_lang=language)

    types.MethodType(build_extension_cpp, build_extension)
```

and then call `edit_build_ext(build_extension)` before `build_extension.run()`

I think this would force C files to be compiled as C++

riddell-stan:
I've lost track of the problem here.

I was thinking we could explore getting numpy install_clib to work. RStan and CmdStan don't do header-only compilation, right? So it seems like the first avenue to explore would be to mirror what they do (which we currently don't).

Hacking distutils seems like a recipe for creating another piece of pystan which is difficult to understand (and hence to maintain). I think that if we can't compile things using Python's un-hacked distutils, we should consider encouraging people to not use Python. (I don't think it will come to this, however, since numpy's install_clib should work)

riddell-stan:
This is outdated, I think.

PR #531, opened 2018-11-13
loop threading test maximum 10 times before failure

ahartikainen:
#### Summary:

Threading summation order is not deterministic, which causes numerical problems.

#### Intended Effect:

Given low probability for the failure (1/100 ~ 1/1000), ten iterations should be enough to succesfully pass the test.

#### How to Verify:

Run tests huge amount of times (follow CI in github)

#### Side Effects:

Tests pass with higher probability

#### Documentation:

None, unless Stan docs don't address this issue (give warning)

#### Copyright and Licensing

Ari Hartikainen

### DIFF
--- a/pystan/tests/test_extra_compile_args.py
+++ b/pystan/tests/test_extra_compile_args.py
@@ -42,7 +42,7 @@ class TestExtraCompileArgs(unittest.TestCase):
                              extra_compile_args=extra_compile_args)
 
     def test_threading_support(self):
-        # Dont test with Windows
+        # Dont test with Windows with MinGW-w64 (GCC)
         if sys.platform.startswith("win"):
             return
         # Set up environmental variable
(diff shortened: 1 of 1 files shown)


sakrejda:
Should probably check with @wds15 since this is one of the issues he pays attention to.

bob-carpenter:
@ahartikainen --- could you elaborate on what the numerical problems are?  Is it just slightly different answers or some kind of bug?  The former should just be dealt with by testing equality up to some precision.  

Also, what were you expecting in the way of Stan documentation on this?  Nothing guarantees exact arithmetic results and that's intentional because of the nature of compilers, CPUs, and floating point.

ahartikainen:
Hi,

What we are testing here is that for a draw, `lp__` equals manually calculated "`lp`".

I was under the impression that this should work (and it works for 1cpu), but we can also change the tests that the values are almost equal.

ahartikainen:
I think we go with this, and if we keep hitting the bug in future too often we can then start to think what is the relative / absolute accuracy we want to test against.

PR #547, opened 2019-01-11
chore: update version number for development

ahartikainen:
Small fix.


riddell-stan:
Thanks. This is actually in the checklist: https://github.com/stan-dev/pystan/blob/develop/doc/release-howto.rst , but I forgot about it.

PR #566, opened 2019-03-20
Update Stan source to tag v2.19.0

riddell-stan:


ahartikainen:
We need to edit also the `stanfit.hpp` for VI to have 3 values before variables in writer

    lp__, log_p__, log_g__, var1, var2, ...

1254

        pystan_sample_writer *sample_writer_ptr
          = sample_writer_factory(&sample_stream,
                                  comment_stream, "# ",
                                  1,
                                  0,
                                  constrained_param_names.size(),
                                  output_samples + 1,
                                  0,
                                  qoi_idx);

to

        pystan_sample_writer *sample_writer_ptr
          = sample_writer_factory(&sample_stream,
                                  comment_stream, "# ",
                                  3,
                                  0,
                                  constrained_param_names.size(),
                                  output_samples + 1,
                                  0,
                                  qoi_idx);

How could we return these samples to user without saving to csv?

pystan/model.py
```diff
@@ -262,8 +262,8 @@ def __init__(self, file=None, charset='utf-8', model_name="anon_model",
             os.path.join(pystan_dir, "stan", "src"),
             os.path.join(pystan_dir, "stan", "lib", "stan_math"),
             os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "eigen_3.3.3"),
-            os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "boost_1.66.0"),
-            os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "sundials_3.1.0", "include"),
+            os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "boost_1.69.0"),
+            os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "sundials_4.1.0", "include"),
```
ahartikainen: Should we add `opencl` folder here too

riddell-stan: Perhaps we should wait to see what RStan does. Cmdstan 2.19 does not add opencl by default.

riddell-stan:
re: vi, how about getting 2.19 out first and then addressing vi? I don't mind making a 2.19.0.1 release.

ahartikainen:
Vi: Sounds fine, but we need to add that 3 to `stanfit.hpp` or vi will throw runtime error.

pystan/model.py
ahartikainen: Sounds good. We really need to test it before release.

setup.py
```diff
@@ -125,7 +125,7 @@ def find_version(*parts):
         extra_compile_args = [
             '/EHsc',
             '-DBOOST_DATE_TIME_NO_LIB',
-            '/std:c++14',
+            '/std:c++1y',
```
ahartikainen: I don't think MSVC has c++1y implemented, so c++14 should be used. (or nothing)

riddell-stan: We only use/support gcc on windows, right? (I'm just following cmdstan and what seemed to work on httpstan)

riddell-stan:
I think the travis failure is due to the old gcc version (4.8.4). I'll try updating things to match the setup that httpstan uses.

We probably also need to update the information about the minimum gcc version. Anyone know what that should be?

setup.py
ahartikainen: Yes. This part is a bit legacy and is here if MSVC starts to work at some point.

ahartikainen:
Minimum GCC for `c++1y` is 4.9.x (I think this was mentioned somewhere in the Discourse)

@wds15 ?

wds15:
Rtools for windows uses Mingw 4.9.3...so that one is used for tests. Unfortunately the threading fix did not make it into the release. So you will still have trouble on windows with this compiler.

ahartikainen:
This is PyStan 2.x, so threading is not enabled by default.

riddell-stan:
We need to move to a newer version of gcc on travis. This may take some serious work. I suggest tackling this separately from updating to v2.19. I'll create a PR in a bit.

ahartikainen:
Could this work

https://github.com/cggos/ptam_cg/issues/1

riddell-stan:
The problem here is, I think, that the newer gcc still uses the older gcc libc++.

riddell-stan:
The `xenial` image solves the problem. If the VI update works then this should be good to go.

ahartikainen:
`test_extra_compile_args` should use `-c++1y`

riddell-stan:
re: windows failures due to long filename (excluded in `MANIFEST.in`)

This seems to be a problem with setuptools's ``bdist_wheel`` ignoring ``MANIFEST.in`` (``sdist`` uses it). There is an issue for it already: https://github.com/pypa/setuptools/issues/511

riddell-stan:
@ahartikainen really long filepaths are breaking the wheel on Windows, e.g., ``build\bdist.win-amd64\wininst\PLATLIB\pystan\stan\lib\stan_math\lib\boost_1.69.0\libs\gil\doc\html\reference\structboost_1_1gil_1_1channel__mapping__type_3_01planar__pixel__reference_3_01_channel_reference60971d1be7c73d78ff725e654349b4f8.html``.

I'd like to just delete every `doc` folder inside `pystan\stan\lib\stan_math\lib`. On Linux I can just do `find pystan/stan/lib/stan_math/lib -type d -iname doc -exec rm -rf "{}" \;`. Any ideas how I can do this in powershell or in windows command-line?

ahartikainen:
I think that will work.

Also this will probably work : `del / *.html`

Also `FOR /d /r "pystan\stan\lib\stan_math\lib" %d IN (doc) DO @IF EXISTS "%d" rd /s /q "%d"`

riddell-stan:
Trying one of your suggestions. If this doesn't work I'm tempted to merge this (since it works fine for Linux and macOS) and ask for help fixing Windows in a separate PR.

This sort of thing is hard for me to work on because I don't have access to a Windows machine.

ahartikainen:
Sounds ok

riddell-stan:
Thanks. FYI the following did not work:

```
FOR /d /r "pystan\stan\lib\stan_math\lib" %d IN (doc) DO @IF EXISTS "%d" rd /s /q "%d"
```

My powershell attempt with

```
- ps: Get-ChildItem pystan -Filter *.html | rm
```

did not error but did not seem to actually delete the files.

ahartikainen:
Oh, I had typo there

    FOR /d /r "pystan\stan\lib\stan_math\lib" %d IN (doc) DO @IF EXIST "%d" rd /s /q "%d"

PR #568, opened 2019-03-25
fix(ci): use newer gcc on travis

riddell-stan:
Newer gcc needed for Stan 2.19.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,14 @@ dist: trusty
 sudo: required
 os:
   - linux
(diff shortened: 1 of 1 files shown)


ahartikainen:
LGTM

riddell-stan:
I'm fairly sure this is still using the gcc 4.x libc++. I'll try switching to xenial in another pr.

riddell-stan:
The xenial version seems to work. Let's use that.

PR #570, opened 2019-03-26
fix(ci): use newer gcc on travis (use xenial)

riddell-stan:
Newer gcc needed for Stan 2.19.


ahartikainen:
LGTM. Now the gcc is 5.4 which should be suitable for `-c++1y`

PR #571, opened 2019-03-28
fix: minor setuptools adjustments

riddell-stan:
Updates files which should be excluded from sdist and removes a stale
reference in setup.py


ahartikainen:
LGTM

PR #574, opened 2019-03-31
chore: update release notes, version number

riddell-stan:
Added a reminder in the release notes: "Reminder: Python 2 is not supported after 2019-12-31. PyStan will phase out testing of Python 2 after this point."

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -6,6 +6,13 @@
  What's New
 ============
 
+v2.19.0.0 (31. Mar 2019)
+========================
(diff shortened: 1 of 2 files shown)


riddell-stan:
@ahartikainen Anything else we should add to the release notes?

ahartikainen:
On a PyStan side, I guess nothing else. 

Info from Stan that changes output:

vb IO: `log_p__` and `log_g__` values are added to sample file (csv) if defined (`sample_file="./out.csv"`).

riddell-stan:
I need to add information about the new gcc requirement. `c++1y` is now a hard requirement with 2.19

PR #586, opened 2019-04-14
External C++

ahartikainen:
#### Summary:

Allow external C++ in Stan program.
If used, a warning is shown that this is an experimental feature.

#### Intended Effect:

Enable external C++ code.

#### How to Verify:

Run example e.g. https://dfm.io/posts/stan-c++/

#### Side Effects:

If not explicitly used, none.

#### Documentation:

Needs update.

#### Copyright and Licensing

@dfm

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -77,7 +77,7 @@ a number of methods.
       the rest for the parameters. Parameters are listed in the same order as `pars`.
 
    .. py:method:: stansummary(pars=None, probs=(0.025, 0.25, 0.5, 0.75, 0.975), digits_summary=2)
-   
+
       Summary statistic table.
       Parameters
       ----------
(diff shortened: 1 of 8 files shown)


riddell-stan:
I'm -1 on this because (1) it adds complexity/features where I think we should be reducing complexity/features and (2) I'm not sure how many people want this feature. I don't think we should add high-level features that benefit far less than 1% of the userbase. These features seem easy to address with a fork and/or documentation.

In this particular case it seems extremely easy for people to maintain a fork since they are already editing C++ code.

ahartikainen:
Maybe, I'm thinking if we move this under kwargs, then one needs to be explicit to use this feature.

The reason for the merge is that this is enabled already in RStan, and this feature is not going to come Stan 2.x codebase.

riddell-stan:
So people are using this a lot with RStan? That would change my opinion.

Another thing I don't like about this is that it seems like the functionality should be supported at the Stan level (not RStan/PyStan-level) in some way if we are going to commit to maintaining it.

ahartikainen:
> So people are using this a lot with RStan? That would change my opinion.

What is a lot? I know some use cases. This also opens the possibility to develop Stan functions from PyStan without recompiling the whole Stan architecture.

> Another thing I don't like about this is that it seems like the functionality should be supported at the Stan level (not RStan/PyStan-level) in some way if we are going to commit to maintaining it.

I agree with this. This is why I don't want to add this as "main" functionality, but as a "hidden" feature for advanced users.

I think Stan3 is going to have a better interface for this?

riddell-stan:
For me, "a lot" means 4 or more people on the discussion forum or in 
Github issues.

(I can't believe there are many people who have the ability to write C++ 
code but who can't patch/fork CmdStan/RStan/PyStan to use their code -- 
but it seems like this patch is precisely for those people, right?)

Adding a warning message for people using this feature sounds like a 
good idea. And we certainly need a test for this. Maybe we could borrow 
RStan's?

ahartikainen:
Yes we need to test this.

RStan has this https://mc-stan.org/rstan/articles/external.html

Do they even test this?

bgoodri:
Yes, the vignette has to compile, which acts as a minimal test. More than 4
people have been referred to it on Discourse
https://discourse.mc-stan.org/search?q=external

riddell-stan:
OK. If RStan does it, I suppose PyStan should too. Is the implementation 
basically the same (same kwarg names)?

ahartikainen:
I have to check out if we can do this only by enabling `allow_undefined` and `#include` structure

ahartikainen:
I think this implementation is more clear than the implementation in RStan external.

```
stan_model(model_code = mc, model_name = "external", allow_undefined = TRUE,
           includes = paste0('\n#include "', 
                             file.path(getwd(), 'fib.hpp'), '"\n'))
```

vs

```
pystan.StanModel(model_code=model_code, allow_undefined=True, includes=["fib.hpp"], include_dirs=["."])
```

bgoodri:
Yeah, I think it is fine to assume / require the .hpp file to be in the
same directory as the .stan file. But in R, the C++ file gets put in a
temporary directory for compiling, which is why we have to do that stuff to
obtain a full path to the .hpp file.

ahartikainen:
`include_dirs` could also have some other location, but I think the safest way is to have the `.hpp` file in the same folder (or subfolder).

We also have `include_paths` which is for Stan `#include stanfile.stan` which kind of needs full path.

ahartikainen:
Let's see, I added a test for C++ function containing grads. I will still add another function that uses Stan autograd to calculate grads.

ahartikainen:
If there are no corrections I will merge this in 24h.

ahartikainen:
Thanks for the implementation @dfm

PR #594, opened 2019-04-30
Ignore optimization errors and return results anyways

blahoink:
#### Summary:

So in ```rstan``` when the ```optimizing``` function errors out it still returns the parameters at the last iteration before the error -- and right now this behavior isn't implemented in ```pystan``` (it throws a runtime error that has to be caught). This change returns parameters from the optimization call regardless of the outcome.

#### Intended Effect:

Get parity with the error behavior of ```optimizing``` in ```rstan```.

#### How to Verify:

Run ```pystan``` with a model/data that fails, and check if you get the parameters returned.

#### Side Effects:

This change will probably make it harder for new users to debug and understand their models, but having this behavior is good for optimization runs that error out simply because the tolerances are not set correctly. Instead of putting the burden on the users to change hyperparameters and re-run what could potentially be a very long computational run -- users can use the returned parameters which could work perfectly well for their purposes.

#### Copyright and Licensing

### DIFF
--- a/pystan/stanfit4model.pyx
+++ b/pystan/stanfit4model.pyx
@@ -401,7 +401,13 @@ def _call_sampler(data, args, pars_oi=None):
         pars_oi_bytes = [n.encode('ascii') for n in pars_oi]
         if len(pars_oi_bytes) != fitptr.param_names_oi().size():
             fitptr.update_param_oi(pars_oi_bytes)
-    ret = fitptr.call_sampler(deref(argsptr), deref(holderptr))
+
+    # Print any runtime exception but don't crash out
+    try:
+        ret = fitptr.call_sampler(deref(argsptr), deref(holderptr))
+    except RuntimeError as e:
+        print('{}: {}'.format(type(e).__name__, str(e)))
+
     holder = _pystanholder_from_stanholder(holderptr)
     # FIXME: rather than fetching the args from the holderptr, we just use
     # the argsptr we passed directly. This is a hack to solve a problem
--- a/pystan/stanfit4model.pyx
+++ b/pystan/stanfit4model.pyx
@@ -403,10 +403,13 @@ def _call_sampler(data, args, pars_oi=None):
             fitptr.update_param_oi(pars_oi_bytes)
 
     # Print any runtime exception but don't crash out
+    call_sampler_error = None
+    ret = 0
     try:
         ret = fitptr.call_sampler(deref(argsptr), deref(holderptr))
     except RuntimeError as e:
         print('{}: {}'.format(type(e).__name__, str(e)))
+        call_sampler_error = e
 
     holder = _pystanholder_from_stanholder(holderptr)
     # FIXME: rather than fetching the args from the holderptr, we just use
@@ -415,7 +418,7 @@ def _call_sampler(data, args, pars_oi=None):
     holder.args = _dict_from_stanargs(argsptr)
     del argsptr
     del fitptr
-    return ret, holder
+    return ret, holder, call_sampler_error
 
 def _split_pars_locs(fnames, pars):
     """Split flatnames to par and location"""
--- a/pystan/model.py
+++ b/pystan/model.py
@@ -459,7 +459,7 @@ class StanModel:
 
         as_vector : boolean, optional
             Indicates an OrderedDict will be returned rather than a nested
-            dictionary with keys 'par' and 'value'.
+            dictionary with keys 'par', 'value', and 'ret'.
 
         Returns
         -------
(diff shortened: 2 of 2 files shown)


ahartikainen:
Is there some key/value added to results if the optimization fails?

blahoink:
No but the return code from the sampler could be easily added to the dictionary that the optimizing function returns. Not sure if we could also add it to the ```as_vector``` result unless we changed the format of the vector.

blahoink:
@ahartikainen I'm now attaching the return code and the error object, but only when as_vector=False. I'm not sure how to integrate those pieces of information... maybe add two fields to the OrderedDict like "__stan_code" and "__stan_error"?

ahartikainen:
I'm now thinking if this will affect libraries using our package (fbprophet, BAMBI).

Will the current stage break the implementations in these libraries.

Should this be an optional feature?

blahoink:
Sure, adding this behavior as a non-default option would still work well. I think that the compiled model should still return the parameters, as then to turn this behavior on would require re-compiling the model -- so the error can be handled accordingly on the python side instead.

Is this acceptable? I can push that change now.

ahartikainen:
@bletham how would this change affect fbprophet?

I still think, it would probably be better to put this functionality under keyword (default = False), because this could break those other packages.

In future editions those packages can test if this functionality is found and then use it if they want.

ahartikainen:
> Sure, adding this behavior as a non-default option would still work well. I think that the compiled model should still return the parameters, as then to turn this behavior on would require re-compiling the model -- so the error can be handled accordingly on the python side instead.

Yes, this could be handled in the python side. Recompiling models is a waste of everyone's time.

blahoink:
Ok, I added this behavior with a flag called ignore_errors that is by default set to False. I've also added the return code and error to the output OrderedDictionary as the keys "__stan_return_code" and "__stan_error"

bletham:
@ahartikainen we do run into errors with optimizer exiting abnormally in some cases with very small datasets, where the model is badly overspecified and the likelihood surface can be expected to be very flat around the optimum. We've found that the Newton optimizer tends to be more robust in handling these very flat optima and so have introduced code to catch the RuntimeError and then retry with Newton: https://github.com/facebook/prophet/blob/master/python/fbprophet/forecaster.py#L1123

If in the future you'd like to change the default here, we can replace that to have ignore_errors=True and look for return_code!=0, at which point a change in default value of ignore_errors would not affect us. But for now defaulting ignore_errors=False would be great.

ahartikainen:
@bletham sounds good. I think default `False` is a good idea. I think many users might just use older pystan with the newer fbprophet because they have older compiler.

ahartikainen:
@blahoink this looks good.

I think we could have a test for testing this behaviour.

@riddell-stan do you have any opinions concerning this?

pystan/stanfit4model.pyx
```diff
@@ -401,15 +401,24 @@ def _call_sampler(data, args, pars_oi=None):
...
+
     holder = _pystanholder_from_stanholder(holderptr)
     # FIXME: rather than fetching the args from the holderptr, we just use
     # the argsptr we passed directly. This is a hack to solve a problem
     # that holder.args gets dropped somewhere in C++.
     holder.args = _dict_from_stanargs(argsptr)
     del argsptr
     del fitptr
-    return ret, holder
+    return ret, holder, call_sampler_error
```
ahartikainen: Other function using this function also need update. Maybe currently reraise the error (`sampling`, `vb`). There could be output to logger in python side before raising the error?

blahoink: Sorry, I'm not sure I understand the problem. I can stop passing the actual error object to the python code to preserve compatibility, and use just the return code instead. Is that the fix you are talking about?

ahartikainen: this now returns 3 parameters, so also sampling and vb are getting these, so I think you can add extra keyword to grab the error and raise it if error was raised.

So 

optimize --> error --> warn in python --> return results (optional)
sampling --> error --> raise error in python
vb --> error --> raise error in python

(I probably need to double check what I'm trying to say)

riddell-stan:
In Python the norm is to use Exceptions when there are errors. Is there information that is desired that is not being added to the Exception? The exception object can hold plenty of information.

blahoink:
Sounds good. Thanks for the comments, I forgot I could add properties to the Exception object.

I'm a little bit worried that the cython code returns another parameter. Users with cached models from a previous version that want this feature turned on will have it not work (or maybe crash?) if they want this behavior. I'll try to do some testing with a cached model from before these changes to see if it produces any bad behavior.

riddell-stan:
Looking into this a bit more.

I think statsmodels raises Exceptions. This seems most natural to me as a Python user.

Looks like sklearn uses `warnings` for failed fits. https://docs.python.org/3.7/library/warnings.html

I can't think of any examples where one is required to check a return code. Having to check a return code strikes me as extremely unpythonic.

ahartikainen:
@mitzimorris what did CmdStanPy decided to do?

mitzimorris:
for CmdStanPy this was hashed out in this PR:  https://github.com/stan-dev/cmdstanpy/pull/75

regarding fbprophet - the final conclusion is:

>  it turns out that the return code 70 will occur rarer that I initially assumed. After detailed inspecting of the lbfgs optimizer code it looks like when there is at least one iteration the return code will in fact be 0.

riddell-stan:
I'm going to close this for now. Seems like some sort of solution was reached.

PR #607, opened 2019-06-03
fix: Fix appveyor link in README

### LINKED ISSUE #604: Appveyor link in README broken
The appveyor account is now "pystan" not "ariddell". There is at least one link (in README) which needs to be updated.

riddell-stan:
Closes #604.

The appveyor docs seem to suggest a slightly different url. I gather it's pending because we have not built the `master` branch on this appveyor account. I suspect it will fix itself when we have the next release.


ahartikainen:
LGTM

PR #609, opened 2019-06-05
Load pickled fit without the model instance

ahartikainen:
#### Summary:

Add functionality to read old fit files, without the correct model instance.

#### Intended Effect:

Read fit, which doesn't have model instance saved.

#### How to Verify:

1. create a fit 
2. pickle only the fit
3. restart python
4. try to load pickle normally
5. load pickle with `pystan.experimental.load_fit`

tests added
~@riddell-stan do you want to add tests for experimental code too?~

#### Side Effects:

None. This is inside experimental module, which is not imported by default.

#### Documentation:

Needs update and simple example. Also needs to have disclaimer, that probably only works with the
original OS. (I have not tested with other OS, so let's test before merge).

#### Copyright and Licensing

Ari Hartikainen

### DIFF
--- a/pystan/experimental/__init__.py
+++ b/pystan/experimental/__init__.py
@@ -1,6 +1,7 @@
-import logging
-from .misc import *
+import logging as _logging
+from .misc import fix_include
+from .pickling_tools import load_fit
 
-logger = logging.getLogger('pystan')
+_logger = _logging.getLogger('pystan')
 
-logger.warning("This submodule contains experimental code, please use with caution")
+_logger.warning("This submodule contains experimental code, please use with caution")
--- a/pystan/experimental/__init__.py
+++ b/pystan/experimental/__init__.py
@@ -1,7 +1,5 @@
-import logging as _logging
+import logging
 from .misc import fix_include
 from .pickling_tools import load_fit
 
-_logger = _logging.getLogger('pystan')
-
-_logger.warning("This submodule contains experimental code, please use with caution")
+logging.getLogger('pystan').warning("This submodule contains experimental code, please use with caution")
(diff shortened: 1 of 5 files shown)


pystan/experimental/__init__.py
```diff
@@ -1,6 +1,7 @@
-import logging
-from .misc import *
+import logging as _logging
+from .misc import fix_include
+from .pickling_tools import load_fit
 
-logger = logging.getLogger('pystan')
+_logger = _logging.getLogger('pystan')
```
riddell-stan: Would ``from pystan import logger`` work?

riddell-stan: `getLogger` is probably right thing to do. I'm just confused about why there are the added underscores. They don't seem necessary.

riddell-stan:
Let me know when this is finalized and I'll review it more closely. In general, looks fine to me. (Although I'm frustrated, of course, that we even need this workaround.)

pystan/experimental/__init__.py
ahartikainen: Yes, I didn't want pollute the namespace because the submodule has only 2 main functions. I can change it back.

riddell-stan: We don't hide `logging` anywhere else. That's what looked weird to me.

If you want to be very economical:
```
import logging
logging.getLogger('pystan').warning('...')
```

ahartikainen:
This is ready (I rebased so tests run again).

It basically creates simple model. I used `-O1` to make compilation faster (`-O0` crashed on Windows).

It reads the module name from pickle file, but if there is a problem with this user can give it by kw.

ahartikainen:
Oh, I still need to add example to docs

ahartikainen:
Yes, this is ready now.

PR #611, opened 2019-06-10
implement inv_metric

ahartikainen:
#### Summary:

Enable user to input initial mass matrix for sampling.
Add function to extract mass matrices from `fit.sim`.
Create a helper function to start sampling from last location with correct settings.

#### Intended Effect:

Use predefined mass matrix. 
Enable user to continue sampling from the last location.

#### How to Verify:

Define and check results

#### Side Effects:

None, if not used

#### Documentation:

Need update

#### Copyright and Licensing

Ari Hartikainen


ahartikainen:
[WIP] : @riddell-stan do have some touch how to handle input-output in `stanfit.hpp` ? Should it get pointer or string?

https://github.com/stan-dev/stan/blob/develop/src/stan/services/sample/hmc_nuts_dense_e.hpp
````
* @param[in] init_inv_metric var context exposing an initial dense
                    inverse Euclidean metric (must be positive definite)

...

int hmc_nuts_dense_e(Model& model, stan::io::var_context& init,
                           stan::io::var_context& init_inv_metric,
                           unsigned int random_seed, unsigned int chain,
                           double init_radius, int num_warmup, int num_samples,
                           int num_thin, bool save_warmup, int refresh,
                           double stepsize, double stepsize_jitter,
                           int max_depth,
                           callbacks::interrupt& interrupt,
                           callbacks::logger& logger,
                           callbacks::writer& init_writer,
                           callbacks::writer& sample_writer,
                           callbacks::writer& diagnostic_writer)

...

Eigen::MatrixXd inv_metric;
        try {
          inv_metric =
            util::read_dense_inv_metric(init_inv_metric, model.num_params_r(),
                                         logger);
          util::validate_dense_inv_metric(inv_metric, logger);
        } catch (const std::domain_error& e) {
          return error_codes::CONFIG;
        }
```

and in `util::read_dense_inv_metric`

https://github.com/stan-dev/stan/blob/develop/src/stan/services/util/read_dense_inv_metric.hpp

```
inline
      Eigen::MatrixXd
      read_dense_inv_metric(stan::io::var_context& init_context,
                            size_t num_params,
                            callbacks::logger& logger) {
        Eigen::MatrixXd inv_metric;
        try {
          init_context.validate_dims(
            "read dense inv metric", "inv_metric", "matrix",
            init_context.to_vec(num_params, num_params));
          std::vector<double> dense_vals =
            init_context.vals_r("inv_metric");
          inv_metric =
            stan::math::to_matrix(dense_vals, num_params, num_params);
        } catch (const std::exception& e) {
          logger.error("Cannot get inverse metric from input file.");
          logger.error("Caught exception: ");
          logger.error(e.what());
          throw std::domain_error("Initialization failure");
        }

        return inv_metric;
      }
```

ahartikainen:
I think there is some problem having `std::string` in the struct.

riddell-stan:
There's an issue tracking this somewhere. My thinking was to wait to see what RStan did. We can copy their strategy and their tests.

This is a challenging task. If you manage to get something working, that would be great.

ahartikainen:
No it kind of has correct values (python side) but keeps seeing 0 (when it should be 1) for metric_flag and also the string is empty (should contain path).

ahartikainen:
Huh, it works now. I still needed to edit one function (dict to args).

I merged those commits now to one. I have done some dictionary kwargs bouncing (because we need to create some external files, and there is some problems with string and struct or something).

I also created small fix for other external filepaths.

I still need to create tests, docs and also some helper functions. (get_mass_matrix etc)

ahartikainen:
I added some convenience functions to extract stepsize and mass-matrix to `pystan.misc` (and wrapper to `StanFit4Model`.

I think creating convenience function to continue sampling from last location could be done in another PR, it needs some logical tests what parameters are needed to and what are not.

Also, currently those functions return lists, but `.sampling` doesn't eat them. For example easiest way to use old mass matrix is to do

    mass_to_control = dict(enumerate(fit.get_mass_matrix())) 

which is then added under `mass_matrix` in `control`.

pystan/model.py
```diff
@@ -719,6 +721,60 @@ def sampling(self, data=None, pars=None, chains=4, iter=2000,
                 logger.warning("`warmup=0` forced with `algorithm=\"Fixed_param\"`.")
             warmup = 0
 
+        # this needs to be done here:
+        # the tempfiles are deleted after sampling has been done
+        # each chain gets unique file
```
ahartikainen: move this under misc

pystan/stan_fit.hpp
```diff
@@ -1094,14 +1158,25 @@ namespace pystan {
 
           if (args.get_ctrl_sampling_metric() == DENSE_E) {
             if (!args.get_ctrl_sampling_adapt_engaged()) {
+              if (args.get_metric_file_flag()) {
               return_code = stan::services::sample
-                ::hmc_static_dense_e(model, *init_context_ptr,
+                ::hmc_static_dense_e(model, *init_context_ptr, *metric_context_ptr,
```
ahartikainen: do we ignore metric if adapt == false?

ahartikainen: Need to checkout what CmdStan does

ahartikainen: https://github.com/stan-dev/cmdstan/blob/develop/src/cmdstan/command.hpp

We don't ignore metric, even if adapt == False.

ahartikainen:
I think this is ready now.  

At least the functionality-side. It's still ok to change parameter names.

ahartikainen:
@riddell-stan are you ok with the userfacing name (`mass_matrix`) and functionality (ndarray / path are saved to tempfile, code run, tempfile deleted; this is because windows has some problems reading same file multiple times/almost same time)

betanalpha:
I strongly recommend inv_metric and not mass_matrix.
The “mass matrix” terminology is burdened by an incomplete
physics analogy and something that I’m trying to move everyone
past

ahartikainen:
Thanks, I think inv_matrix is good name.

ahartikainen:
I updated: `mass_matrix` -> `inv_metric`

Also updated the examples in docs:

- use seed (pseudo-continue)
- example showing `arviz.concat`

ArviZ concat functionality is currently only in GitHub version (master branch). The next release is in August (end of GSoC). I can handle the issues if there are any.
- solution: `pip install git+https://github.com/arviz-devs/arviz`

ahartikainen:
I will merge in 24h if there are no comments.

PR #616, opened 2019-06-26
Update inv_metric behaviour

ahartikainen:
#### Summary:

Update docstring
Handle edge cases (no samples)

#### Intended Effect:

Better behavior for the function

#### Copyright and Licensing

Ari Hartikainen

### DIFF
--- a/doc/hmc_euclidian_metric.rst
+++ b/doc/hmc_euclidian_metric.rst
@@ -59,7 +59,10 @@ To make this reproducible, user can manually set seed for each fit.
     # increment seed by 1
     seed2 = seed + 1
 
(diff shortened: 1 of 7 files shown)


ahartikainen:
There is something weird going with Adaptation info and stepsize. I'm investigating if that is only a local glitch.

(It always returns stepsize=1)

ahartikainen:
Ok, `adapt_engaged` needs to be `False`

ahartikainen:
I will still add `warmup` option for the `last_position` function, so user can "continue" from the warmup if wanted. 

Also there is some bug with `warmup`==`iter`, so need to check if that is something we want to fix.

ahartikainen:
I will merge this and evaluate sampling problems in another PR

PR #626, opened 2019-08-07
Update Stan to v2.20.0

riddell-stan:


ahartikainen:
LGTM

ahartikainen:
There is some segfault

Is threading tests failing? Maybe disable it for now, or do we want to wait, until it is fixed?

https://github.com/stan-dev/pystan/blob/develop/pystan/tests/test_extra_compile_args.py#L43

    def test_threading_support(self):
        # Dont test with Windows with MinGW-w64 (GCC)
        if sys.platform.startswith("win"):
            return
        # Set up environmental variable
        os.environ['STAN_NUM_THREADS'] = "2"
        # Enable threading
        extra_compile_args = ['-pthread', '-DSTAN_THREADS']
        stan_code = """
        functions {
          vector bl_glm(vector mu_sigma, vector beta,
                        real[] x, int[] y) {
            vector[2] mu = mu_sigma[1:2];
            vector[2] sigma = mu_sigma[3:4];
            real lp = normal_lpdf(beta | mu, sigma);
            real ll = bernoulli_logit_lpmf(y | beta[1] + beta[2] * to_vector(x));
            return [lp + ll]';
          }
        }
        data {
          int<lower = 0> K;
          int<lower = 0> N;
          vector[N] x;
          int<lower = 0, upper = 1> y[N];
        }
        transformed data {
          int<lower = 0> J = N / K;
          real x_r[K, J];
          int<lower = 0, upper = 1> x_i[K, J];
          {
            int pos = 1;
            for (k in 1:K) {
              int end = pos + J - 1;
              x_r[k] = to_array_1d(x[pos:end]);
              x_i[k] = y[pos:end];
              pos += J;
            }
          }
        }
        parameters {
          vector[2] beta[K];
          vector[2] mu;
          vector<lower=0>[2] sigma;
        }
        model {
          mu ~ normal(0, 2);
          sigma ~ normal(0, 2);
          target += sum(map_rect(bl_glm, append_row(mu, sigma),
                                 beta, x_r, x_i));
        }
        """
        stan_data = dict(
            K = 4,
            N = 12,
            x = [1.204, -0.573, -1.35, -1.157,
                 -1.29, 0.515, 1.496, 0.918,
                 0.517, 1.092, -0.485, -2.157],
            y = [1, 1, 0, 0, 1, 1, 0, 0, 0, 1, 1, 1]
        )
        stan_model = pystan.StanModel(
            model_code=stan_code,
            extra_compile_args=extra_compile_args
        )

ahartikainen:
Weird thing is, that clang samples fine with threading, and gcc fails.

Is there some flag that gcc needs?

riddell-stan:
Oh, I didn't realize we had a test with `STAN_THREADS`. I'm sure it's the same problem we're discussing here: https://discourse.mc-stan.org/t/example-of-use-of-autodiffstacksingleton-stan-threads-related/10037

I'll see if I can get clang working locally. Maybe newer versions of gcc work? I don't think gcc 7 works but we could try gcc 8? Do you have that installed?

riddell-stan:
I think the simplest thing is to drop support for STAN_THREADS in PyStan 2 on Linux and ask people to use PyStan 3 (where it's not optional).

ahartikainen:
I tested pystan-next with updated httpstan 

This segfaulted with gcc 7, 8, 9 and clang 8

https://github.com/stan-dev/pystan-next/blob/master/tests/test_corr_matrix.py

ahartikainen:
Ok, more testing. Ubuntu with default gcc 7, threading works (tested with pytest).

Is this segfaulting deterministic or random?

riddell-stan:
(I'll submit a patch that fixes things for clang but not gcc on httpstan -- but let's discuss that in httpstan, not here)

Just narrowly regarding pystan 2, I feel like we have two options in the very short-term.

1. Drop support for gcc on Linux in 2.20
2. Drop support for STAN_THREADS in 2.20 (and remove the test)
3. Do not release 2.20, wait until a future release of stan-math that allows STAN_THREADS and gcc.

I'm in favor of 2. What do you think? (Are there options I missed?)

ahartikainen:
I'm fine with 2. Threading was / is experimental feature (on pystan).

riddell-stan:
Not sure what's happening with the Python 3.7 tests.

ahartikainen:
My guess is that cython / distutils or something else has just updated and has some bug with 3.7.

pystan/tests/test_extra_compile_args.py
```diff
@@ -39,89 +39,3 @@ def test_extra_compile_args_failure(self):
         with assertRaisesRegex(distutils.errors.CompileError, 'failed with exit status'):
             pystan.StanModel(model_code=model_code, model_name="normal1",
                              extra_compile_args=extra_compile_args)
-
-    def test_threading_support(self):
```
ahartikainen: Can we add

    import unittest

    @unittest.skip("temporarily disabled")
    def test_threading_support

which causes it not to be run with other tests, but let's us test is locally?

https://docs.python.org/2/library/unittest.html#unittest.skip

riddell-stan:
The 3.7 failures are due to a conda bug, right? If that's the case, I'm OK with ignoring them and merging this (and releasing pystan 2.20).

ahartikainen:
LGTM

PR #660, opened 2019-10-26
ci: Do not test on or build wheels for Python 3.5

riddell-stan:
Any objections? This only applies going forward. 2.19.1 has 3.5 wheels.


ahartikainen:
Numpy drops py35 in January 2020 (?)

PR #661, opened 2019-10-26
ci: Do not build 32bit Windows wheels

riddell-stan:
Does anyone depend on 32bit wheels?


ahartikainen:
Apparently there are multiple 32bit Windows users in corporate environments.

Can we somehow provide external wheels, like repo, which has only 32bit wheels?

PR #667, opened 2019-10-28
Rebase cvodes on v2.19.1.1

riddell-stan:
```
git rebase v2.19.1.1 cvodes
```

I think this avoids merge commits and keeps history cleaner. I could be wrong about this.


ahartikainen:
Great, yes rebase is the correct method here.

PR #687, opened 2020-02-20
Updated installation instructions for Vanilla Python users

Lplenka:
I faced an issue while executing these commands:
``import pystan``
``model_code = 'parameters {real y;} model {y ~ normal(0,1);}'``
``model = pystan.StanModel(model_code=model_code)``

There phrase "type ``export CC=gcc_linux-64`` and ``export CXX=g++_linux-64``" is kind of misleading for vanilla python users. Hence, added these lines.

#### Copyright and Licensing


ahartikainen (approved):
Thanks for the instructions

riddell-stan:
These steps aren't necessary on a recent version of, say, Ubuntu or Arch or any distribution which ships with a recent gcc.

Perhaps start with instructions for people with 2019 or 2020 vintage distributions and then have a separate section for earlier distributions. (We don't want people installing gcc 7 when they already have gcc 9!)

Lplenka:
I agree with @riddell-stan. But not everyone in the developer community will be using the latest Linux OS or Anaconda.

The installation instructions are supposed to be more generic and for a wider range of audience. Current instructions are mostly for Anaconda Users.

Well, the instructions I added will work for gcc version greater than 7 too. I just did not want vanilla python users people to confuse installation instructions for Anaconda users with theirs.

doc/installation_beginner.rst
```diff
@@ -139,7 +139,11 @@ You should see two compilers like ``x86_64-apple-darwin13.4.0-clang`` and ``x86_
 
 - type ``export CC=x86_64-apple-darwin13.4.0-clang`` and ``export CXX=x86_64-apple-darwin13.4.0-clang++``
 
+If not using Anaconda (in Linux OS)
+---------------------
 
+- You need to install gcc-7 using this `reference <https://forums.linuxmint.com/viewtopic.php?p=1465419&sid=72631feb0aa7bc551f199c606aecf256#p1465419>`_.
```
ahartikainen, line 145: I think we don't need to be specific about the `gcc-7`, but this example is good.

Maybe we rephrase it so users need gcc (4.9.x+? with C++14 support).

doc/installation_beginner.rst
```diff
@@ -139,7 +139,11 @@ You should see two compilers like ``x86_64-apple-darwin13.4.0-clang`` and ``x86_
...
+- Set-up your path to include folder containing compilers (e.g. In case ``which gcc-7`` returns ``/usr/bin/gcc-7``) then add ``export CC=/usr/bin/gcc-7`` and ``export CXX=/usr/bin/g++-7`` to your ~/.bashrc file.
```
ahartikainen, line 146: Maybe we can add example (is the following correct?)

    echo 'export CC=/usr/bin/gcc-7' >> ~/.bashrc
    echo 'export CXX=/usr/bin/g++-7' >> ~/.bashrc

ahartikainen (commented):
Hi, thanks for being patient. Let's finalize this PR.

doc/installation_beginner.rst
Lplenka: I agree we don't need to be specific about version of gcc.

doc/installation_beginner.rst
Lplenka: This is true for "gcc-7" since CC and CXX variables should be available for successful installation of Pystan. This will work seamlessly for vanilla python users.

riddell-stan:
Closing this before marking this repo as archived. PyStan 3 will be released on March 25, 2021.

PR #692, opened 2020-04-01
Update/stan 2.22.1 math 3.1.1

ahartikainen:
#### Summary:

Updates `stan` source to 2.22.1 and `stan_math` to 3.0.0

Fixes:
- Update var_context
- build and link to TBB (`stan-math/lib/tbb`)
    - TBB is build at setup step and distributed with `.whl`, it is not "installed".
    - Use `rpath` for TBB linking on Linux/macOS
    - Add `TBB` to PATH in load_module function on Windows
- `make` as a build requirement (`mingw32-make` on Windows)
- update stan paths (`_chains.pyx`)
- `-D_REENTRANT` added as a `CXXFLAG`
- Stan threading is now always on, still uses multiprocessing for `n_jobs`. (STAN_NUM_THREADS for multiple threads)
- Threading works on Windows with mingw-w64 compiler. Added testing for threading on Windows.

#### Intended Effect:

Update Stan.

#### How to Verify:

Run tests.

#### Side Effects:

Extra handling for TBB.

#### Documentation:

Threading docs updated.

#### Copyright and Licensing

Ari Hartikainen

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 build/
 dist/
 doc/_build
+pystan/tests/cached_models
 
 __pycache__
 *.a
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@ doc/_build
 __pycache__
 *.a
 *.so
+*.dll
+*.pyd
 *.cpp
 *.pyc
 *.egg-info
--- a/pystan/__init__.py
+++ b/pystan/__init__.py
@@ -5,10 +5,9 @@
 # License. See LICENSE for a text of the license.
 #-----------------------------------------------------------------------------
 import logging
-import platform
 
 from pystan.api import stanc, stan
-from pystan.misc import read_rdump, stan_rdump, stansummary, add_libtbb_path
+from pystan.misc import read_rdump, stan_rdump, stansummary
 from pystan.diagnostics import check_hmc_diagnostics
 from pystan.model import StanModel
 from pystan.lookup import lookup
@@ -17,13 +16,6 @@ logger = logging.getLogger('pystan')
 logger.addHandler(logging.NullHandler())
 if len(logger.handlers) == 1:
     logging.basicConfig(level=logging.INFO)
-
-if platform.system() == "Windows":
-    add_libtbb_path()
-
-# clean imports
-del add_libtbb_path
-del platform
 
 # following PEP 386
 # See also https://docs.openstack.org/pbr/latest/user/semver.html
--- a/pystan/__init__.py
+++ b/pystan/__init__.py
@@ -19,4 +19,4 @@ if len(logger.handlers) == 1:
 
 # following PEP 386
 # See also https://docs.openstack.org/pbr/latest/user/semver.html
-__version__ = '2.22.1.0'
+__version__ = '2.22.1.0dev'
--- a/pystan/__init__.py
+++ b/pystan/__init__.py
@@ -7,7 +7,7 @@
 import logging
 
 from pystan.api import stanc, stan
-from pystan.misc import read_rdump, stan_rdump, stansummary
+from pystan.misc import read_rdump, stan_rdump, stansummary, add_libtbb_path
 from pystan.diagnostics import check_hmc_diagnostics
 from pystan.model import StanModel
 from pystan.lookup import lookup
@@ -16,6 +16,9 @@ logger = logging.getLogger('pystan')
 logger.addHandler(logging.NullHandler())
 if len(logger.handlers) == 1:
     logging.basicConfig(level=logging.INFO)
+
+add_libtbb_path()
+del add_libtbb_path
 
 # following PEP 386
 # See also https://docs.openstack.org/pbr/latest/user/semver.html
--- a/pystan/__init__.py
+++ b/pystan/__init__.py
@@ -5,6 +5,7 @@
 # License. See LICENSE for a text of the license.
 #-----------------------------------------------------------------------------
 import logging
+import platform
 
 from pystan.api import stanc, stan
 from pystan.misc import read_rdump, stan_rdump, stansummary, add_libtbb_path
@@ -17,8 +18,12 @@ logger.addHandler(logging.NullHandler())
 if len(logger.handlers) == 1:
     logging.basicConfig(level=logging.INFO)
 
-add_libtbb_path()
+if platform.system() == "Windows":
+    add_libtbb_path()
+
+# clean imports
 del add_libtbb_path
+del platform
 
 # following PEP 386
 # See also https://docs.openstack.org/pbr/latest/user/semver.html
--- a/pystan/__init__.py
+++ b/pystan/__init__.py
@@ -22,4 +22,4 @@ del add_libtbb_path
 
 # following PEP 386
 # See also https://docs.openstack.org/pbr/latest/user/semver.html
-__version__ = '2.22.0.0'
+__version__ = '2.22.1.0'
--- a/pystan/__init__.py
+++ b/pystan/__init__.py
@@ -19,4 +19,4 @@ if len(logger.handlers) == 1:
 
 # following PEP 386
 # See also https://docs.openstack.org/pbr/latest/user/semver.html
-__version__ = '2.19.1.2dev'
+__version__ = '2.22.0.0'
--- a/pystan/experimental/pickling_tools.py
+++ b/pystan/experimental/pickling_tools.py
@@ -16,7 +16,7 @@ import sys
 import shutil
 
 from pystan._compat import PY2
-from pystan.model import load_module, build_tbb
+from pystan.model import load_module
 
 logger = logging.getLogger('pystan')
 
--- a/pystan/experimental/pickling_tools.py
+++ b/pystan/experimental/pickling_tools.py
@@ -64,7 +64,7 @@ def create_fake_model(module_name):
         os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "eigen_3.3.3"),
         os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "boost_1.72.0"),
         os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "sundials_4.1.0", "include"),
-        os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "tbb_2019_U8", "include"),
+        os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "tbb", "include"),
         np.get_include(),
     ]
 
--- a/pystan/experimental/pickling_tools.py
+++ b/pystan/experimental/pickling_tools.py
@@ -16,7 +16,7 @@ import sys
 import shutil
 
 from pystan._compat import PY2
-from pystan.model import load_module
+from pystan.model import load_module, build_tbb
 
 logger = logging.getLogger('pystan')
 
@@ -32,6 +32,15 @@ def create_fake_model(module_name):
     StanModel
         Dummy StanModel object with specific module name.
     """
+    tbb_dir = os.getenv("STAN_TBB")
+    if tbb_dir is None:
+        tbb_dir = os.path.join(os.path.dirname(__file__), '..', 'stan', 'lib', 'stan_math', 'lib','tbb')
+        tbb_dir = os.path.abspath(tbb_dir)
+        if not os.path.exists(tbb_dir):
+            build_tbb()
+    else:
+        tbb_dir = os.path.abspath(tbb_dir)
+
     # reverse engineer the name
     model_name, nonce = module_name.replace("stanfit4", "").rsplit("_", 1)
     # create minimal code
@@ -57,8 +66,9 @@ def create_fake_model(module_name):
         os.path.join(pystan_dir, "stan", "src"),
         os.path.join(pystan_dir, "stan", "lib", "stan_math"),
         os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "eigen_3.3.3"),
-        os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "boost_1.69.0"),
+        os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "boost_1.72.0"),
         os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "sundials_4.1.0", "include"),
+        os.path.join(pystan_dir, "stan", "lib", "stan_math", "lib", "tbb_2019_U8", "include"),
         np.get_include(),
     ]
 
@@ -107,7 +117,10 @@ def create_fake_model(module_name):
                 "-D_hypot=hypot",
                 "-pthread",
                 "-fexceptions",
+                '-DSTAN_THREADS',
+                '-D_REENTRANT',
             ] + extra_compile_args
+        extra_link_args = []
     else:
         # linux or macOS
         extra_compile_args = [
@@ -116,14 +129,21 @@ def create_fake_model(module_name):
             '-Wno-unused-function',
             '-Wno-uninitialized',
             '-std=c++1y',
+            '-DSTAN_THREADS',
+            '-D_REENTRANT',
         ] + extra_compile_args
+        extra_link_args = ['-Wl,-rpath,{}'.format(os.path.abspath(tbb_dir))]
 
     extension = Extension(name=module_name,
                           language="c++",
                           sources=[pyx_file],
                           define_macros=stan_macros,
                           include_dirs=include_dirs,
-                          extra_compile_args=extra_compile_args)
+                          libraries=["tbb"],
+                          library_dirs=[tbb_dir],
+                          extra_compile_args=extra_compile_args,
+                          extra_link_args=extra_link_args
+                          )
 
     cython_include_dirs = ['.', pystan_dir]
 
--- a/pystan/experimental/pickling_tools.py
+++ b/pystan/experimental/pickling_tools.py
@@ -32,9 +32,7 @@ def create_fake_model(module_name):
     StanModel
         Dummy StanModel object with specific module name.
     """
-    tbb_dir = os.getenv("STAN_TBB")
-    if tbb_dir is None:
-        tbb_dir = os.path.join(os.path.dirname(__file__), '..', 'stan', 'lib', 'stan_math', 'lib','tbb')
+    tbb_dir = os.path.join(os.path.dirname(__file__), '..', 'stan', 'lib', 'stan_math', 'lib','tbb')
     tbb_dir = os.path.abspath(tbb_dir)
 
     # reverse engineer the name
@@ -170,7 +168,7 @@ def get_module_name(path, open_func=None, open_kwargs=None):
     module_name = ""
     with open_func(path, **open_kwargs) as f:
         # scan first few bytes
-        for i in range(500):
+        for i in range(10000):
             byte, = f.read(1)
             if not PY2:
                 byte = chr(byte)
--- a/pystan/experimental/pickling_tools.py
+++ b/pystan/experimental/pickling_tools.py
@@ -36,8 +36,6 @@ def create_fake_model(module_name):
     if tbb_dir is None:
         tbb_dir = os.path.join(os.path.dirname(__file__), '..', 'stan', 'lib', 'stan_math', 'lib','tbb')
         tbb_dir = os.path.abspath(tbb_dir)
-        if not os.path.exists(tbb_dir):
-            build_tbb()
     else:
         tbb_dir = os.path.abspath(tbb_dir)
 
--- a/pystan/experimental/pickling_tools.py
+++ b/pystan/experimental/pickling_tools.py
@@ -35,9 +35,7 @@ def create_fake_model(module_name):
     tbb_dir = os.getenv("STAN_TBB")
     if tbb_dir is None:
         tbb_dir = os.path.join(os.path.dirname(__file__), '..', 'stan', 'lib', 'stan_math', 'lib','tbb')
-        tbb_dir = os.path.abspath(tbb_dir)
-    else:
-        tbb_dir = os.path.abspath(tbb_dir)
+    tbb_dir = os.path.abspath(tbb_dir)
 
     # reverse engineer the name
     model_name, nonce = module_name.replace("stanfit4", "").rsplit("_", 1)
--- a/pystan/misc.py
+++ b/pystan/misc.py
@@ -463,7 +463,7 @@ def _config_argss(chains, iter, warmup, thin,
         inits_specified = True
     if not inits_specified and isinstance(init, Callable):
         ## test if function takes argument named "chain_id"
-        if "chain_id" in inspect.getargspec(init).args:
+        if "chain_id" in inspect.getfullargspec(init).args:
             inits = [init(chain_id=id) for id in chain_id]
         else:
             inits = [init()] * chains
--- a/pystan/misc.py
+++ b/pystan/misc.py
@@ -463,7 +463,8 @@ def _config_argss(chains, iter, warmup, thin,
         inits_specified = True
     if not inits_specified and isinstance(init, Callable):
         ## test if function takes argument named "chain_id"
-        if "chain_id" in inspect.getfullargspec(init).args:
+        getfullargspec = inspect.getfullargspec if hasattr(inspect, "getfullargspec") else inspect.getargspec
+        if "chain_id" in getfullargspec(init).args:
             inits = [init(chain_id=id) for id in chain_id]
         else:
             inits = [init()] * chains
--- a/pystan/misc.py
+++ b/pystan/misc.py
@@ -1492,7 +1492,7 @@ def add_libtbb_path():
             )
         )
 
-def build_stan_tbb():
+def build_stan_tbb(use_debug=False):
     """Build tbb."""
     stan_math_lib = os.path.abspath(os.path.join(os.path.dirname(__file__), 'stan', 'lib', 'stan_math', 'lib'))
 
@@ -1519,6 +1519,9 @@ def build_stan_tbb():
     tbb_release = os.path.join(stan_math_lib, "tbb_release")
     tbb_dir = os.path.join(stan_math_lib, "tbb")
 
+    if use_debug:
+        tbb_release, tbb_debug = tbb_debug, tbb_release
+
     if not os.path.exists(tbb_dir):
         os.makedirs(tbb_dir)
 
@@ -1527,7 +1530,8 @@ def build_stan_tbb():
 
     for name in os.listdir(tbb_release):
         srcname = os.path.join(tbb_release, name)
-        shutil.move(srcname, tbb_dir)
+        dstname = os.path.join(tbb_dir, name)
+        shutil.move(srcname, dstname)
 
     if os.path.exists(tbb_release):
         shutil.rmtree(tbb_release)
--- a/pystan/misc.py
+++ b/pystan/misc.py
@@ -1476,62 +1476,11 @@ def get_last_position(fit, warmup=False):
     return positions
 
 def add_libtbb_path():
-    """Add libtbb to PATH on Windows."""
-    if platform.system() == "Windows":
-        libtbb = os.getenv('STAN_TBB')
-        if libtbb is None:
-            libtbb = os.path.abspath(os.path.join(
-                os.path.dirname(__file__), 'stan', 'lib', 'stan_math', 'lib', 'tbb'
-            ))
+    """Add libtbb to PATH."""
+    libtbb = os.path.abspath(os.path.join(
+        os.path.dirname(__file__), 'stan', 'lib', 'stan_math', 'lib', 'tbb'
+    ))
 
-        os.environ['PATH'] = ';'.join(
-            list(
-                OrderedDict.fromkeys(
-                    [libtbb] + os.getenv('PATH', '').split(';')
-                )
-            )
-        )
-
-def build_stan_tbb(use_debug=False):
-    """Build tbb."""
-    stan_math_lib = os.path.abspath(os.path.join(os.path.dirname(__file__), 'stan', 'lib', 'stan_math', 'lib'))
-
-    make = os.getenv('MAKE', 'make' if platform.system() != 'Windows' else 'mingw32-make')
-    cmd = [make]
-
-    tbb_root = os.path.join(stan_math_lib, 'tbb_2019_U8').replace("\\", "/")
-    tbb_src = os.path.join(tbb_root, 'src').replace("\\", "/")
-
-    cmd.extend(['-C', tbb_root])
-    cmd.append('tbb_build_dir={}'.format(stan_math_lib))
-    cmd.append('tbb_build_prefix=tbb')
-    cmd.append('tbb_root={}'.format(tbb_root))
-
-    cmd.append('stdver=c++14')
-
-    compiler = os.getenv('TBB_COMPILER', 'gcc')
-    cmd.append('compiler={}'.format(compiler))
-
-    cwd = os.path.abspath(os.path.dirname(__file__))
-    subprocess.check_call(cmd, cwd=cwd)
-
-    tbb_debug = os.path.join(stan_math_lib, "tbb_debug")
-    tbb_release = os.path.join(stan_math_lib, "tbb_release")
-    tbb_dir = os.path.join(stan_math_lib, "tbb")
-
-    if use_debug:
-        tbb_release, tbb_debug = tbb_debug, tbb_release
-
-    if not os.path.exists(tbb_dir):
-        os.makedirs(tbb_dir)
-
-    if os.path.exists(tbb_debug):
-        shutil.rmtree(tbb_debug)
-
-    for name in os.listdir(tbb_release):
-        srcname = os.path.join(tbb_release, name)
-        dstname = os.path.join(tbb_dir, name)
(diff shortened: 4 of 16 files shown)


ahartikainen:
Ok this basically works.

~~- Multiprocessing does not work.~~
    - ~~I tested quickly MultiThreading, but it kept crashing (on windows).~~

- ~~Pickling does not work.~~
- Setup needs to be fixed, so tbb is build at the setup step.
    - I think we could do the following at the pre-install step (for wheel) or in post-install for the normal build (?). Not sure what is the correct way so the binaries would go to the correct destination.

https://stackoverflow.com/questions/20288711/post-install-script-with-python-setuptools

Wheel (pre-built binaries):
We have a couple of options.
1. Prebuilt, and pack in the wheel as is.
2. Audit wheel etc.

I think the first should work, because we can add "helper" scripts to add the specific folder the PATH / LDFLAGS ( I think rpath thing means, we don't really need LDFLAGS). For Windows, this is the easiest way. I'm not certain how rpath thing works, if the location for libtbb.so changes, but is still in PATH or LDFLAGS. 

Option 1 would be the easiest way to handle this, because tbb is not really installed anywhere when we call `make`.

https://stackoverflow.com/questions/23916186/how-to-include-external-library-with-python-wheel-package

@ariddell what do you think?

Then we have some other issues. Stan lib does not have JSON files anymore (they have been moved under CmdStan), so I'm not sure how can we handle that, when we have small wrapper for JSON input files for inverse metric. Is there already a way to read in python and send the information to Stan, or do we drop the support for JSON inverse file (I think Rdump should still work).
Or do we copy that C++ JSON files? cc @mitzimorris 

I don't know how big the operation is to use precompiled model files. Is that something what could be precompiled and then just injected to `Extension` or do we need a lot of some custom building? Does that work the same way as Sundials setup works (they are "precompiled" and then send as `.o` files to the `Extension`? cc @bob-carpenter 

What (new) flags do we need to use? @wds15 

    -DSTAN_THREADS  # is this still needed?
    -D_REENTRANT  # why??
    -Wl,-rpath,path/to/tbb' # linux / osx only
    STAN_NUM_THREADS # == n_jobs?

Also, if I fail to cc right person, please correct me

This PR is for ~2.21~ 2.22(?), but I don't think after this there should be anything big why we could not also update 2.22 and 2.23 (?). 2.22 did have some small bug, so I think releasing 2.21 is a good idea.

edit. I think this is for 2.22, but should we also do a release for 2.21?

pystan/py_var_context.hpp
```diff
@@ -168,7 +167,7 @@ namespace pystan {
        *
        * @param names Vector to store the list of names in.
        */
-      virtual void names_r(std::vector<std::string>& names) const {
```
ahartikainen: I think these were removed by an accident

setup.py
```diff
@@ -101,8 +103,9 @@ def find_version(*parts):
 stan_include_dirs = ['pystan/stan/src',
                      'pystan/stan/lib/stan_math/',
                      'pystan/stan/lib/stan_math/lib/eigen_3.3.3',
-                     'pystan/stan/lib/stan_math/lib/boost_1.69.0',
-                     'pystan/stan/lib/stan_math/lib/sundials_4.1.0/include']
+                     'pystan/stan/lib/stan_math/lib/boost_1.72.0',
+                     'pystan/stan/lib/stan_math/lib/sundials_4.1.0/include',
+                     'pystan/stan/lib/stan_math/lib/tbb_2019_U8/include']
```
ahartikainen: Not sure if tbb folder needed here

wds15:
Stan threads is to make the ad tape tls  aware. The reentrant is for the right log gamma. The rpath stuff hard coded the path to the shared library of the tbb. This does. Not work on windows. On windows the tbb.dll must be in the system path or in the same directory as the model binary.

ahartikainen:
> Stan threads is to make the ad tape tls aware. The reentrant is for the right log gamma. 

Ok, that sounds good.

I'm not sure do we even need `-DSTAN_THREADS`, I will test this.

> The rpath stuff hard coded the path to the shared library of the tbb. This does. Not work on windows. 

Yes, I implemented that only for Linux / macOS

> On windows the tbb.dll must be in the system path or in the same directory as the model binary.

I added a function that reads `STAN_TBB` environmental variable and if not found, add the default location to the PATH. And this is done at import step. User can also manually call `pystan.misc.add_libtbb_path` which does this check again and if not found, uses the default. And this is a Windows only function.

Is there a way to overwrite the rpath location for compiled model on Linux / macOS. For example user uses pickle and saves the file and opens it in another computer or later with different installation location.

wds15:
You do need Stan threads.

Mac and Linux have utilities to change rpath...but the design assumes you start the model on a machine where the Stan install path is stable. You don’t have to deal with rpath if you ensure that the linker finds the tbb.

ahartikainen:
Ok, pickling btw works, I needed to update that one experimental function that creates pseudo model.

So `-DSTAN_THREADS` should be always on, right?

wds15:
It works without Stan threads...but you mentioned multi threading, that’s why I think you should turn it on. You only pay 1-2% performance penalty for some models. For most models there is no cost due to it. So you gain a lot of robustness and users can use map rect with threads. To me there is no price to pay and you gain a lot of robustness and within chain parallelism...so just turn it on is my judgement.

ahartikainen:
Oh, I was tripped by the rookie mistake with multiprocessing and Windows. I needed to call a function under the `__name__ == "__main__"` block.

So multiprocessing works as it should.

ahartikainen:
~~Ok, this is ready. Only need to update docs.~~

ahartikainen:
~I think we can add disclaimer that JSON input for `inverse_metric` does not currently work. Or even add a custom transformer for it.~

edit. We have not even supported the json version apparently, so nothing really changes.

pystan/model.py
```diff
@@ -54,20 +53,28 @@ def load_module(module_name, module_path):
...
-def _map_parallel(function, args, n_jobs):
+def _map_parallel(function, args, n_jobs, n_threads=None):
     """multiprocessing.Pool(processors=n_jobs).map with some error checking"""
     # Following the error checking found in joblib
     multiprocessing = int(os.environ.get('JOBLIB_MULTIPROCESSING', 1)) or None
     if multiprocessing:
         try:
             import multiprocessing
             import multiprocessing.pool
+            if n_jobs < 1:
```
riddell-stan: I think we calculate the number of jobs elsewhere. Or joblib interprets -1 as num cpu.

pystan/model.py
```diff
@@ -54,20 +53,28 @@ def load_module(module_name, module_path):
...
+                n_jobs = multiprocessing.cpu_count()
         except ImportError:
             multiprocessing = None
-        if sys.platform.startswith("win") and PY2:
-            msg = "Multiprocessing is not supported on Windows with Python 2.X. Setting n_jobs=1"
+            n_jobs = 1
```
riddell-stan: We should raise an exception or somehow abort if we cannot use the user's arguments. Overriding a user's arguments is unusual.

pystan/model.py
```diff
@@ -54,20 +53,28 @@ def load_module(module_name, module_path):
...
+        if (platform.system() == 'Windows') and PY2:
+            msg = 'Multiprocessing is not supported on Windows with Python 2.X. Setting n_jobs=1'
             logger.warning(msg)
             n_jobs = 1
+
+    old_n_threads = os.getenv('STAN_NUM_THREADS')
```
riddell-stan: Being able to change behavior via arguments or environment variables seems to add complexity. It's also unusual for a Python package like PyStan to change its behavior in response to environment variables. I don't think scikit-learn or pandas or numpy allow much customization of their behavior via environment variables.

pystan/model.py
```diff
@@ -226,6 +238,12 @@ def __init__(self, file=None, charset='utf-8', model_name="anon_model",
                  obfuscate_model_name=True, extra_compile_args=None,
                  allow_undefined=False, include_dirs=None, includes=None):
 
+        tbb_dir = os.getenv("STAN_TBB")
```
riddell-stan: We don't let users set the path to boost with an environment variable. Is there an argument for allowing it for tbb specifically?

pystan/__init__.py
```diff
@@ -17,6 +17,9 @@
 if len(logger.handlers) == 1:
     logging.basicConfig(level=logging.INFO)
 
+add_libtbb_path()
```
riddell-stan: I would add the platform check before this line (and remove it from the body of add_libtbb_path). That way the reader of this code knows what the line is for.

pystan/model.py
ahartikainen: Yes, this was left here by an accident. I will remove this.

riddell-stan:
re: number of threads. I'd suggest we pick a sensible default. Is there a reason to pick something higher than the number of CPUs?

Python picks # cpus times five for ThreadPoolExecutor. But there they are assuming IO-heavy work. See https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor

ahartikainen:
I didn't pick anything, because user needs to play with n_jobs + n_threads and it is currently only for map_rect.

So maybe default to cpu_count --> physical cores + virtual? (`multiprocessing.cpu_count`)

This is much easier in pystan3 where TBB can handle everything.

setup.py
```diff
@@ -179,6 +224,12 @@ def find_version(*parts):
 package_data_pats = ['*.hpp', '*.pxd', '*.pyx', 'tests/data/*.csv',
                      'tests/data/*.stan', 'lookuptable/*.txt']
 
+# Build tbb before setup if needed
+tbb_path = os.path.join(os.path.dirname(__file__), 'pystan', 'stan', 'lib', 'stan_math', 'lib', 'tbb')
+print(tbb_path)
```
ahartikainen: This is debug line, should be removed

setup.py
```diff
@@ -179,6 +224,12 @@ def find_version(*parts):
...
+if not os.path.exists(tbb_path):
```
ahartikainen: Not sure if we should test against the folder or files.

This is simpler.

riddell-stan:
re: num_threads, I think defaulting to the number of cpu cores is fine. Those who need to set it to something else can do so.

ahartikainen:
Ok, I used cpu count for the default (n_threads == None and STAN_NUM_THREADS is not defined).

I updated threading information and What's new.

For general user, nothing on PyStan side changes.

The compilation time might have been increased a bit, I have not clocked it, but it might be something that users will notice.

After we merge this, I think we should make a release and then try to see how can we incorporate the latest pre-compiled model parts which will decrease the compilation time.

wds15:
Do you want me to comment things?

ahartikainen:
> Do you want me to comment things?

Yes, you are welcome to comment. It's always good to have comments and opinions because after a while I start to become blind to my mistakes/approaches.

wds15:
Are you also going to support building the stan services in advance to speedup compilation times? That's very interesting for users as it cuts quite a bit down the compile times. However, it requires some care with linking. See https://github.com/stan-dev/math/issues/1750#issuecomment-594707130

ahartikainen:
Maybe. Not sure if we should first implement it in pystan3 or here.

If there is some known workflow from setup --> model + linking --> sampling then I can't think any reason why not.

Do we have a list of things we can do?

- precompile stan model parts
- precompile services

This pr "precompiles" TBB, so user won't notice it. And `cvodes` branch compiles Sundials each time it compiles model, but basically that step could be done at setup time, as we do for TBB now.

wds15:
You should definitely include the sundials stuff in the pre-built step here. Maybe do the services pre-compilation another time, but users will love it (>50% compile time reduction).

ahartikainen:
Yes, that is something we can do. Just in another PR.

setup.py
ahartikainen: I think I will keep this here, it doesn't hurt anything.

ahartikainen:
@riddell-stan  do you want me to clean up the commits or do we squash them?

riddell-stan:
I'm a fan of squashing.

There is one unresolved review comment, the one involving ` add_libtbb_path()`. Do you mind addressing that?

pystan/misc.py
```diff
@@ -1472,3 +1474,64 @@ def get_last_position(fit, warmup=False):
         extract_pos = {key : values[draw_location, i] for key, values in extracted.items()}
         positions.append(extract_pos)
     return positions
+
+def add_libtbb_path():
+    """Add libtbb to PATH on Windows."""
+    if platform.system() == "Windows":
+        libtbb = os.getenv('STAN_TBB')
```
riddell-stan: One more item: why are we allowing the user to set STAN_TBB here? We don't do it for boost or eigen. It adds extra lines of code too.

riddell-stan: I double checked. We don't allow configuration via environment variables anywhere. The only exception is joblib, but I think that's a reasonable exception since it follows a convention established for that library.

Also, I think the standard is to use os.environ.get, not os.getenv, see https://stackoverflow.com/a/16924553

pystan/misc.py
```diff
@@ -1472,3 +1474,64 @@ def get_last_position(fit, warmup=False):
...
+    tbb_src = os.path.join(tbb_root, 'src').replace("\\", "/")
+
+    cmd.extend(['-C', tbb_root])
+    cmd.append('tbb_build_dir={}'.format(stan_math_lib))
+    cmd.append('tbb_build_prefix=tbb')
+    cmd.append('tbb_root={}'.format(tbb_root))
+
+    cmd.append('stdver=c++14')
+
+    compiler = os.getenv('TBB_COMPILER', 'gcc')
```
riddell-stan: If we distribute binary wheels virtually nobody is going to be settings this, right?

riddell-stan:
I added a few more comments. Thanks again for figuring this out. Translating this to work with httpstan will be easy. 👍

ahartikainen:
Sorry, I don't see any comments? Are they still pending and not send?

riddell-stan (changes requested):

riddell-stan:
You're right. I forgot to submit them.

pystan/model.py
ahartikainen: yeah, this was my debug line, I removed it

pystan/model.py
ahartikainen: That is true. Maybe we should not use it. I will remove these.

pystan/model.py
ahartikainen: TBB can be shared by many programs, so user might want to use system tbb files and then distribute resources between multiple programs.

That's why I set it so that user can select what tbb folder to use.

pystan/__init__.py
ahartikainen: Good call, I will change this

wds15:
BTW, is it really necessary to introduce Stan_tbb? Pystan would be the only interface with this config variable ... and the default to use the tbb which you distribute should work.

pystan/misc.py
ahartikainen: I will change os.environ.get

TBB comment as above --> TBB can be used by many programs, so user might want to use system tbb

pystan/misc.py
ahartikainen: I think we probably can remove this function. It was meant for users who want to use debug version etc, but tbh I think those users can use raw `make`. So let's remove this function from misc.

ahartikainen:
> BTW, is it really necessary to introduce Stan_tbb? Pystan would be the only interface with this config variable ... and the default to use the tbb which you distribute should work.

It is not. Should we let users to decide what TBB to use?

wds15:
I don’t quite see a point to let users choose which tbb to use... in case the system has one then it would be good to use that, but we have never enabled things like this so far.

For math we have cxxflags_tbb and the like defined in the makefile which can be overridden if needed. I doubt that people will use this. Everyone will just use the default is my intuition.

ahartikainen:
Ok, I will remove those and let's use the defaults

riddell-stan:
You're right that TBB (and boost, etc) can be shared by other programs. 
The practice right now is to "vendor" everything so pystan uses its own 
copy of each library.

We can change this but we should change it for all libraries, not just tbb.

ahartikainen:
Do we need a way to turn of threading? Os there a flag for that?

wds15:
There are very few models which run 2% or so slower without threading turned on ( and these models are very fast anyways). Thus, I would just turn threading on to save users from confusion...but you can choose not to define Stan threads.

ahartikainen:
I mean, currently threading is always on (hardcoded), but does nthreads (STAN_NUM_THREADS) = 1 turn it of (or 0)

wds15:
Stan num threads is only runtime control. Setting this to 1 will simply force the tbb to use only 1 thread.

ahartikainen:
Is there still some changes? I don't see anything, but the interface says that there is still one change requested.

wds15:
There is one thing I wanted to bring up here: Right now the `STAN_NUM_THREADS` is meant as a global thing... i.e. in the context of cmdstan it makes sense to set this thing globally as this means it does apply to the chain you start. However, if I am not mistaken, then you are running the models in a threadpool itself... is that correct?

Say, the user can start model1-chain1, model1-chain2 & model2-chain1, etc... and all of that at the same time, right? If that is the case, then it would make sense to tell the TBB to use `STAN_NUM_THREADS` in each of these contexts if I am not mistaken (which it likely doesn't do right now).

You are hitting possibly against the boundaries of stan services library which really should allow to deal with threads... but there would be means to deal with this.

ahartikainen:
PyStan (this repo) uses process pool. So each model+chain is run in individual process. All these are then linked to same TBB, right? So I think TBB distributes resources (threads), which are shared between the different processes, or does it only handle threads in one process, so the process threads are not "touching" other process threads.

PyStan3 (~pystan-next~ httpstan repo) uses thread pool.

wds15:
Ok... if you run these chains as different processes, then what you do for now is fine. The TBB defs for how many threads are used is "static" which is adequate per process.

ahartikainen:
Should we move tbb include folder under tbb and then remove tbb source?

wds15:
I am not sure if I follow what you intent to do with moving the TBB around, but that's probably specific to how PyStan organises libraries.

In any case - I saw that you show an example on how to use threading based on `map_rect`. That's fine, I would just recommend to bump this to `reduce_sum` once you go to 2.23. The new `reduce_sum` is  really easy for users to use and we should encourage people to go with that.

pystan/__init__.py
```diff
@@ -17,6 +18,13 @@
 if len(logger.handlers) == 1:
     logging.basicConfig(level=logging.INFO)
 
+if platform.system() == "Windows":
+    add_libtbb_path()
+
+# clean imports
```
riddell-stan: This practice of deleting imports is new to me. Is this something other libraries do? What's the justification?

edit: nevermind, I didn't realize this was in `__init__.py`. Why does add_libtbb_path need to be added here and not in models.py?

pystan/experimental/pickling_tools.py
```diff
@@ -32,6 +32,11 @@ def create_fake_model(module_name):
     StanModel
         Dummy StanModel object with specific module name.
     """
+    tbb_dir = os.getenv("STAN_TBB")
```
riddell-stan: I feel like we should be consistent here. Either we allow all libraries to set their paths with environment variables, or we let none of them do so (my preference). And even if we introduce this new practice, we should do it in a different pull request.

edit: fixed reference

riddell-stan (changes requested):

pystan/__init__.py
ahartikainen: It is a runtime dependency, so it should be defined at runtime.

Calling it in model.sampling etc could work, but this way it is always defined when we import the lib.

pystan/experimental/pickling_tools.py
ahartikainen: True, I forgot to fix this code. It should not need STAN_TBB call.

riddell-stan:
Here my thinking about why it is not a runtime dependency: a user can 
use pystan to interact with a pickled fit and pickled model -- at no 
point do they need to have their path patched.

I think it would be more elegant to have this called only when it is 
needed. I will not insist on this though. If you're happy with this, 
I'll accept it.

ahartikainen:
It will be added only for the current Python process, so it is not a global thing.

We could add code to call it in sampling function, but do we need it for log_prob calls too?

riddell-stan:
I haven't encountered a situation exactly like this before. While I maintain that it doesn't need to be called in `__init__.py`, I'm not sure where it belongs. Since it's serving the same role as an rpath setting, telling an extension module where to find a shared library, it needs to be called before the tbb shared library gets used.

Can it be added to the top of `stanfit4model.pyx`?

ahartikainen:
I think it can, cython can call python functions.

Will it be called also when one loads the pickle?

riddell-stan:
Anytime the model is imported the path should be modified. Unpickling 
the model and using it will import the model, I think.

ahartikainen:
I will modify and tests should tell us if this works or not.

ahartikainen:
`stanfit4model` thing did not work, but interestingly adding that to `load_module` function works. Even pickle works as expected.

pystan/model.py
```diff
@@ -64,10 +66,17 @@ def _map_parallel(function, args, n_jobs):
             import multiprocessing.pool
         except ImportError:
             multiprocessing = None
-        if sys.platform.startswith("win") and PY2:
-            msg = "Multiprocessing is not supported on Windows with Python 2.X. Setting n_jobs=1"
+        if (platform.system() == 'Windows') and PY2:
+            msg = 'Multiprocessing is not supported on Windows with Python 2.X. Setting n_jobs=1'
```
riddell-stan, line 70: I still think this is really unpythonic. We should raise an exception, not modify the user's arguments. Did you forget to change this?

Also, isn't this a Python 3 and a Python 2 issue? Why is there a PY2 check?

pystan/model.py
```diff
@@ -88,6 +97,9 @@ def _map_parallel(function, args, n_jobs):
             pool.join()
     else:
         map_result = list(map(function, args))
+    # reset STAN_NUM_THREADS
```
riddell-stan: Did you forget to remove this?

pystan/model.py
```diff
@@ -64,10 +66,17 @@ def _map_parallel(function, args, n_jobs):
             import multiprocessing.pool
         except ImportError:
             multiprocessing = None
-        if sys.platform.startswith("win") and PY2:
-            msg = "Multiprocessing is not supported on Windows with Python 2.X. Setting n_jobs=1"
+        if (platform.system() == 'Windows') and PY2:
+            msg = 'Multiprocessing is not supported on Windows with Python 2.X. Setting n_jobs=1'
             logger.warning(msg)
             n_jobs = 1
+
```
riddell-stan: If we are going to introduce the practice of configuring things via environment variables, we should do it in a different pull request.

pystan/misc.py
```diff
@@ -462,7 +462,8 @@ def _config_argss(chains, iter, warmup, thin,
         inits_specified = True
     if not inits_specified and isinstance(init, Callable):
         ## test if function takes argument named "chain_id"
-        if "chain_id" in inspect.getargspec(init).args:
+        getfullargspec = inspect.getfullargspec if hasattr(inspect, "getfullargspec") else inspect.getargspec
```
riddell-stan: Why is this being changed? Is this a compatibility thing?

pystan/experimental/pickling_tools.py
```diff
@@ -154,7 +168,7 @@ def get_module_name(path, open_func=None, open_kwargs=None):
     module_name = ""
     with open_func(path, **open_kwargs) as f:
         # scan first few bytes
-        for i in range(500):
+        for i in range(10000):
```
riddell-stan, line 171: Does this fix some bug?

pystan/__init__.py
```diff
@@ -19,4 +19,4 @@
 
 # following PEP 386
 # See also https://docs.openstack.org/pbr/latest/user/semver.html
-__version__ = '2.19.1.2dev'
+__version__ = '2.22.1.0'
```
riddell-stan: Should be 2.22.1.0dev

.gitignore
```diff
@@ -1,10 +1,13 @@
 build/
 dist/
 doc/_build
+pystan/tests/cached_models
```
riddell-stan, line 4: Is this for your local user environment? Does everyone need this?

riddell-stan (changes requested):
Almost there. Just a few minor points and one or two things that should be fixed.

.gitignore
ahartikainen: Yeah, I mean if you do

    pip install -e .

And then run the tests, you get cached models folder which is only a local folder. (We reuse some models)

pystan/__init__.py
ahartikainen: True

wds15:
BTW... 2.23 is about there - likely this week. As it includes reduce_sum you may want to jump straight to that release? Either directly in this PR or as a new PR prior to releasing this.

pystan/experimental/pickling_tools.py
ahartikainen: if user saves stan str model first. It is not perfect, but user can add the name manually

pystan/misc.py
ahartikainen: Yes, python 3 vs python 2. Assuming we start to move pystan3 soon, then this wont break stuff in future.

pystan/model.py
ahartikainen: This is only a python 2 issue. Python 3 users can use `__name__ == '__main__'` trick

ahartikainen: And only on windows

pystan/model.py
ahartikainen: What should we change in stanfit to enable threading? @wds15?

pystan/model.py
ahartikainen: I left it there because we edit env var.

Not really happy about it either. Can we edit stanfit.hpp to enable some specific flag?

ahartikainen:
I might have a fix for mac multiprocessing issue, but I think I need to do it in another PR

pystan/model.py
wds15: I would recommend you to always compile with STAN_THREADS being defined.

You can stick to configuring things with the environment variable, but there are better options.

The crux is a bit that the thread configuration should better be handled in the stan services,  but that's still some time until we get that, I am afraid.

The better approach from my view is to use so-called task arena's from the TBB. These can be configured to allow some maximal concurrency and you run things within them. That's not too hard to do and it makes the environment variables obsolete which is what I would prefer as an interface dev.

ahartikainen: `STAN_THREADS` is called implemented as always on.

How would we automate task arena for a user? Do we need to start that first and then close it? Is it running as compiled C++ --> do we need to wrap it in python?

I would not want to add tbb-python as a requirement for pystan.

wds15: I agree in that this should be part of the stan services... but that will likely take a moment.

So yes, you would have to instantiate a C++ object and run things with it... it's simple:

https://software.intel.com/en-us/node/506359

this is what should be done in the stan services layer somewhere, but we don't have it yet.

pystan/misc.py
riddell-stan: Please correct me if I'm wrong. It seems like this has nothing to do with updating pystan to Stan 2.22.1.

I think it would be better if these sorts of things were raised for discussion as issues or distinct pull requests.

pystan/model.py
riddell-stan: I didn't realize Python 2 users couldn't do the __main__ trick.

riddell-stan: Sorry, I thought you were introducing the behavior of modifying the user's arguments in this pull request. I gather that was already there. 

But again, please correct me if I'm wrong, it doesn't seem like the changes here have anything to do with the pull request title.

pystan/model.py
riddell-stan: @wds15 and @ahartikainen , how about waiting until this is all appropriately added to the stan services layer. We don't need to introduce bleeding edge stuff into pystan now that there are alternatives like cmdstanpy.

It would be a big accomplishment simply to get pystan updated to the current stan release. We don't need to do everything in this pull request.

@wds15 , is there an issue where we can track progress towards a more user-friendly way of dealing with the num_threads setting?

pystan/misc.py
ahartikainen: True, let's discuss this in another PR /issue.

pystan/model.py
ahartikainen: Yes, these are not directly related to 2.22, I was just updating the call. Let's fo this in another PR.

ahartikainen: Yes, these are not directly related to 2.22, I was just updating the call. Let's fo this in another PR.

pystan/model.py
ahartikainen: That is a good idea.

wds15: Sure. No there is no issue for that with the Stan repo. It will be a big effort to pull that off as it will likely impact a lot of the api. The issue is that knowledge on threads is with another set of people than that with people having a deep overview of the stan services...but that’s a different story.

pystan/model.py
riddell-stan: > BTW... 2.23 is about there - likely this week. As it includes reduce_sum you may want to jump straight to that release? Either directly in this PR or as a new PR prior to releasing this.

I think it's good to take it slow. Aren't there other changes we would need to make for 2.23? Is using stanc3 required?

pystan/model.py
riddell-stan: Having an issue to point to would be useful.

Can we hope that a sensible default might be chosen for people who don't have num threads set? That would be helpful in the short run. And I imagine it would be easier to implement in Stan services.

wds15: In the Short run we have Stan num Threads.

pystan/model.py
ahartikainen: This is now reset, I just left ( " --> ' )

pystan/model.py
ahartikainen: I have now removed `n_threads` so user can still manually define `NUM_STAN_THREADS`.

`-DSTAN_THREADS` is still always used.

pystan/model.py
```diff
@@ -339,10 +347,13 @@ def __init__(self, file=None, charset='utf-8', model_name="anon_model",
...
                     '-Wno-uninitialized',
                     '-std=c++1y',
-                    "-D_hypot=hypot",
-                    "-pthread",
-                    "-fexceptions",
+                    '-D_hypot=hypot',
+                    '-pthread',
+                    '-fexceptions',
+                    '-DSTAN_THREADS',
+                    '-D_REENTRANT',
```
riddell-stan, line 354: Maybe add a comment about _REENTRANT? Is this a windows requirement?

riddell-stan:
Looks good. Definitely squash the commits into one or a smaller number of commits.

I'd also recommend building a wheel yourself, creating a completely independent venv, installing the wheel into that, and running tests. With httpstan (and rpath) I found that in some cases the wheel wasn't as environment/path independent as I thought it was.

riddell-stan:
@wds15 @ahartikainen What happens if ``STAN_NUM_THREADS`` is not defined? Is `1` the default? I assume everything works, right?

riddell-stan (approved):

pystan/model.py
ahartikainen: https://stackoverflow.com/questions/2601753/what-is-the-reentrant-flag

I think this is related to TBB (all platforms).

ahartikainen:
```
- STAN_NUM_THREADS is not defined => num_threads=1
 * - STAN_NUM_THREADS is positive => num_threads is set to the
 *   specified number
 * - STAN_NUM_THREADS is set to -1 => num_threads is the number of
 *   available cores on the machine
 * - STAN_NUM_THREADS < -1, STAN_NUM_THREADS = 0 or STAN_NUM_THREADS is
 *   not numeric => throws an exception
```

https://github.com/stan-dev/math/blob/d46916e7df0cde11365afcaae2a9a4770482daf8/stan/math/prim/core/init_threadpool_tbb.hpp

ahartikainen:
> Looks good. Definitely squash the commits into one or a smaller number of commits.

Do I squash manually or do we do `Squash and merge` when merging.

> I'd also recommend building a wheel yourself, creating a completely independent venv, installing the wheel into that, and running tests. With httpstan (and rpath) I found that in some cases the wheel wasn't as environment/path independent as I thought it was.

Good idea

pystan/model.py
ahartikainen: @wds15 do you know why we need _REENTRANT? It was somewhere mentioned at some point.

ahartikainen:
Works when installed from a wheel in a clean environment. (Tested on Windows and Linux).

(We don't test this on CI?)

riddell-stan:
We test aspects of this with the CI. We don't test everything. For 
example, we don't test in a setting where the machine which built the 
wheel has a different libstdc++ than the machine which is compiling stan 
programs and sampling from them.

pystan/model.py
ahartikainen: https://github.com/stan-dev/math/commit/1ef61d46cc4c46784a24e47691217d559fea28e3

riddell-stan:
I wish this had been squashed before it was merged. 70 commits and only 16 files changed seems excessive.

ahartikainen:
It was squash merged --> 1 commit only in develop.

